PR #11, opened 2022-02-06
feat: Add marker drag event callback to Marker composable

### LINKED ISSUE #10: Marker drag event callback
I want to get marker drag event callback and manipulate marker-related work.

So, `Marker` composable have to have dragging callback parameter like `onMarkerDrag` `onMarkerDragStart` `onMarkerDragEnd` similar with `onClick` and `onInfoWindow~`

It is an alternative to `OnMarkerDragListener` of Android view version.

chibatching:
To observe dragged marker object, I added `onMakerDrag` callback lambda to `Marker` composable

---------

Fixes #10  🦕


jpoehnelt:
@chibatching Please take a moment to fill out [this short survey](https://forms.gle/wYuzN6UibnFj7ZL29). Thank you!

_This is an automated message, feel free to ignore._

maps-compose/src/main/java/com/google/maps/android/compose/Marker.kt
```diff
@@ -55,6 +55,12 @@ class MarkerDragState {
      */
     var dragState: DragState by mutableStateOf(DragState.END)
         internal set
+
+    /**
+     * Object of dragged marker.
```
arriolac, line 60:
```suggestion
     * The dragged marker
```

arriolac (commented):
One minor suggestion. Would you mind also adding usage of this to the sample app to help folks how to use this?

chibatching:
Ok, I'll add it as soon as possible!

maps-compose/src/main/java/com/google/maps/android/compose/Marker.kt
```diff
@@ -46,6 +46,9 @@ enum class DragState {
     START, DRAG, END
 }
 
+@Immutable
+class MarkerWrapper(val value: Marker? = null)
```
chibatching, line 50: I made a wrapper class to add `Immutable` annotation. Please let me know if you have any good ideas.

chibatching:
I made changes to the sample code and fixes the observing marker state problem.

app/src/main/java/com/google/maps/android/compose/MapSampleActivity.kt
```diff
@@ -90,10 +93,15 @@ class MapSampleActivity : ComponentActivity() {
...
 private fun GoogleMapView(modifier: Modifier, onMapLoaded: () -> Unit) {
     val singapore = LatLng(1.35, 103.87)
+    var markerPosition by remember { mutableStateOf(singapore) }
     // Observing and controlling the camera's state can be done with a CameraPositionState
     val cameraPositionState = rememberCameraPositionState {
         position = CameraPosition.fromLatLngZoom(singapore, 11f)
     }
+    val markerDragState = rememberMarkerDragState()
+    if (markerDragState.dragState == DragState.END) {
+        markerDragState.marker.value?.let { markerPosition = it.position }
```
arriolac, line 103: Hm this is an interesting consequence of this feature. The initially set position state will no longer accurate when dragged which is already true for the existing version of the library. Having to do the above is not ideal, the `Marker` composable should handle this for you.

I think a more intuitive design would be to emulate how `CameraPositionState` works. So something like the following:

```kotlin
class MarkerPositionState(
    position: LatLng = LatLng(0.0f, 0.0f)
) {
    var position: LatLng by mutableStateOf(position)

    var dragState: DragState by mutableStateOf(DragState.END)
        internal set
}
```

This means updating the `position` property to accept `MarkerPositionState` and removal of the `markerDragState` as the `DragState` would now be accessed via `MarkerPositionState`.

Some pros/cons of this:
  * pro: `position` would always be consistent as the marker is dragged
  * con: would cause a breaking change
  * con: may have some odd behaviors when the same state is passed to multiple marker instances

Other considerations:
  * note that only the `position` is exposed and not the full `Marker` object. I'm not convinced that it should be exposed unless it is useful for the `tag` property to be accessed while the `Marker` is dragged.

arriolac (commented):
I have a concern with how this is currently implemented, and may need to be redesigned due having two sources of truth for the marker's position. I proposed a new design, have a look.

arriolac:
@chibatching after reading through Compose's [API guidelines](https://github.com/androidx/androidx/blob/androidx-main/compose/docs/compose-api-guidelines.md#prefer-stateless-and-controlled-composable-functions) which states "Prefer stateless and controlled @Composable functions", the callback-based approach makes more sense here over what I proposed in https://github.com/googlemaps/android-maps-compose/pull/11#discussion_r802972785 However, I think there are a few behavior and API changes needed. These will be breaking changes and necessitate a major version update.

* `markerDragState` should be removed from the `Marker` composable. Instead, it should accept a callback called `onMarkerDrag` of type `(Marker, DragState) -> Unit`.
* it is expected that implementations of `onMarkerDrag` should modify a `MutableState<LatLng>` position (i.e. the position provided in the `Marker.position` composable). Failure to do so won't actually drag the marker. This way, the marker position state is fully controlled by the caller of `Marker`. Note that the existing implementation violates this assumption.

chibatching:
I fixed conflicts

app/src/main/java/com/google/maps/android/compose/MapSampleActivity.kt
```diff
@@ -127,9 +130,16 @@ private fun GoogleMapView(modifier: Modifier, onMapLoaded: () -> Unit) {
             false
         }
         MarkerInfoWindowContent(
-            position = singapore,
+            position = markerPositionState,
             title = "Zoom in has been tapped $ticker times.",
             onClick = markerClick,
+            draggable = true,
+            onMarkerDrag = { marker, dragState ->
+                markerPositionState = marker.position
```
arriolac, line 138: If this line was removed `markerPositionState` will be out-of-sync with the marker's actual position. I don't think there's really anyway around this given how the marker's position is managed by the map. Hoisting a state object as I mentioned in https://github.com/googlemaps/android-maps-compose/pull/11#discussion_r802972785 may be the only appropriate solution here.

chibatching: Ah, I'm sorry. I misunderstood and I agree with you. I'll fix it.

maps-compose/src/main/java/com/google/maps/android/compose/Marker.kt
```diff
@@ -86,15 +96,15 @@ fun rememberMarkerDragState(): MarkerDragState = remember {
...
  * @param onClick a lambda invoked when the marker is clicked
  * @param onInfoWindowClick a lambda invoked when the marker's info window is clicked
  * @param onInfoWindowClose a lambda invoked when the marker's info window is closed
  * @param onInfoWindowLongClick a lambda invoked when the marker's info window is long clicked
+ * @param onMarkerDrag a lambda invoked while the marker is dragged
  */
 @Composable
 fun Marker(
-    position: LatLng,
+    positionState: MarkerPositionState = rememberMarkerPositionState(),
```
chibatching, line 107: I replaced raw position with position state completely. But should I remain position parameter or make overload?

arriolac:
@chibatching I've given this a bit more taught and while "prefer stateless and controlled `@composable` functions" is the recommendation, taking this approach will not behave well in Maps Compose since ultimately it is an interop library of the Maps SDK. And, the Maps SDK internally manages drag state of a marker. So, trying to combat that by designing a callback-based approach as I mentioned in https://github.com/googlemaps/android-maps-compose/pull/11#issuecomment-1035586376 would cause marker dragging to be jerky. This would definitely be the preferred API though if Maps Compose were a Compose native implementation of Maps SDK.

Given this, the right approach here would be to use [hoisted state types](https://github.com/androidx/androidx/blob/androidx-main/compose/docs/compose-api-guidelines.md#hoisted-state-types) as I mentioned in https://github.com/googlemaps/android-maps-compose/pull/11#discussion_r802972785. Apologies for the circuitous decisions around the correct API though I think it's important to weigh the pros/cons here.

Let me know if you think you can update this PR with the following changes. If not, I'm happy to take it from here. Really appreciate your contributions thus far.

chibatching:
@arriolac Please don't worry about it. I support your deeply considered decision.
I am still new to building applications using Compose, so this was a great learning experience.

I fixed my PR. If you have any problems, please don't hesitate to tell me!

maps-compose/src/main/java/com/google/maps/android/compose/Marker.kt
arriolac: An overload would be ideal, but I think it would be a maintenance burden to support polymorphic types for `position` and `positionState`. So, I'm in favor of replacing `position` and `markerDragState` here.

arriolac (commented):
This PR looks good to me! Since this is a breaking change, please change the base branch to `maps-compose-2.0.x`. I'll create a `SNAPSHOT` of `2.0` so folks can try out this change. 
I'm considering adding another hoisted state object that contains `MarkerPositionState` (like `MarkerState`) which also contains controls for showing/hiding info window before promoting  this change to 2.0.

chibatching:
@arriolac Thank you for your review!!
And also I changed the base branch to `maps-compose-2.0.x`.

arriolac (commented):
Just one more modification and then should be good to merge.

arriolac (approved):
LGTM. Thanks @chibatching

arriolac:
FYI `MarkerPositionState` was renamed to `MarkerState` in #69 since that object now has functionality other than controlling the marker's position (i.e. showing/hiding the info window).

PR #21, opened 2022-02-08
feat: Add parameter on Marker to control info window visibility

### LINKED ISSUE #16: Displaying info window as soon as a marker is added
Currently it is not possible to display the info window when adding a marker without the user clicking on the marker.
For reference, on the existing Maps API we can utilize `marker.showInfoWindow()` 

A Boolean on the `Marker` Composable to immediately show the info window.

scavazzini:
This pull request adds a new parameter on `Marker` composable to control the info window visibility.

---

Fixes #16 🦕


maps-compose/src/main/java/com/google/maps/android/compose/Marker.kt
```diff
@@ -41,6 +41,11 @@ internal class MarkerNode(
     }
 }
 
+@Immutable
+enum class InfoWindowVisibility {
+    VISIBLE, HIDDEN
```
arriolac, line 46:
```suggestion
    SHOWN, HIDDEN
```
I recommend that the chosen verbs here match that of the original API.

maps-compose/src/main/java/com/google/maps/android/compose/Marker.kt
```diff
@@ -148,6 +155,13 @@ fun Marker(
             set(flat) { this.marker.isFlat = it }
             set(icon) { this.marker.setIcon(it) }
             set(infoWindowAnchor) { this.marker.setInfoWindowAnchor(it.x, it.y) }
+            set(infoWindowVisibility) {
```
arriolac, line 158: This doesn't quite handle all the different edge cases. As currently written, this only handles updates to the Marker composable and not initialization.

arriolac (changes requested):
Thanks for the pull request! It would be great to add this control to info windows. I think there are a couple of other considerations here that can be improved in the design. 

I imagine the info window's visibility to be a `State` similar to how `MarkerDragState` is with the exception of the state being controllable from the outside. As a consequence of this, the following scenarios also need to be handled:
* Update the state when an info window is shown/hidden. This will depend on the work that I'm doing in #22 and you'll have to rebase your changes.
* `onInfoWindowClose` is moot as a result and should be removed.

...also, to test all of this, the sample app should be updated so users know how to use this new info window state pattern.

I recognize this is a lot to ask for an open source contributor but wanted to get my thoughts down here and see if you're open to help with the changes. I can get to this feature as well once I wrap up #22 .

scavazzini:
Hello, thanks for your feedback!

I implemented your suggestions and made info window's visibility as an state that can be controlled from outside.

It works when I set the state's initialValue as HIDDEN, but not when is SHOWN.
Looks like the correct solution is not that simple as I did, right?

arriolac:
Looks like this got closed since the PR was against my branch. Please reopen against `main` and we can revisit. Thanks!

scavazzini:
> Looks like this got closed since the PR was against my branch. Please reopen against `main` and we can revisit. Thanks!

I'm not sure if my implementation was on the right path, so I don't know if is valid to reopen to try again.

arriolac:
@scavazzini yes I think there are a couple of edge cases that need to be handled by your PR. Happy to collab on those scenarios if you are open to contributing. Otherwise, I'd like to get to this issue after #10

PR #22, opened 2022-02-08
feat: Custom info window.

### LINKED ISSUE #3: Implement custom info windows
Custom info windows cannot be implemented at the moment.

`setInfoWindowAdapter` should be exposed on the `GoogleMap` or `Marker` composable. The implementation should allow using Compose for setting the content or info window vs. having to use Views.

arriolac:
## Summary

Support adding a custom info window on markers. 

This change introduces elements `MarkerInfoWindow` and `MarkerInfoWindowContent` which both accept a `content` property for customizing the entire info window, or just the contents of the default info window, respectively. If no customizations are needed, `Marker` should continue to be used.

For more information on custom info windows, see [Custom Info Windows](https://developers.google.com/maps/documentation/android-sdk/infowindows) and [InfoWindowAdapter](https://developers.google.com/android/reference/com/google/android/gms/maps/GoogleMap.InfoWindowAdapter).

## Example Usage

No customization:
```kotlin
Marker(
    // ...
)
```

Customizing the contents of the default info window:

```kotlin
MarkerInfoWindowContent(
  // ...
) { marker ->
    Text(marker.title ?: "Default Marker Title", color = Color.Red)
}
```

Customizing the entire info window:

```kotlin
MarkerInfoWindow(
    // ...
) { marker ->
    Column {
       Text(marker.title ?: "Default Marker Title", color = Color.Red)
       Text(marker.snippet ?: "Default Marker Snippet", color = Color.Red)
    }
  }
} 
```

## Alternatives Considered

* Providing 2 separate composable lambda properties, `infoWindow` and `infoWindowContent`, in `Marker`

This was the original design which is quite similar to the API in `InfoWindowAdapter`. I opted against this design since usage can be quite unexpected—only one lambda expression will ever be invoked (i.e. if `infoWindow` is provided, `infoWindowContent` will be ignored). You can see the documentation for [getInfoContents](https://developers.google.com/android/reference/com/google/android/gms/maps/GoogleMap.InfoWindowAdapter#public-abstract-view-getinfocontents-marker-marker) to learn more about this behavior. To improve upon this, only the composable should only accept one property, `content`, which is the preferred behavior of just customizing the contents of the info window.

* Using a sealed class to provide customizations in a `Marker`. Only one customization (info window, or contents) can be provided in the `Marker` composable element. See [354995a](https://github.com/googlemaps/android-maps-compose/pull/22/commits/354995a355a903e0dd23ed21254f0aa02ac2dcf0) for implementation

This design pattern breaks Compose's API guidelines. See [Compose UI layouts](https://github.com/androidx/androidx/blob/androidx-main/compose/docs/compose-api-guidelines.md#compose-ui-layouts).

Fixes #3 🦕


maps-compose/src/main/java/com/google/maps/android/compose/Marker.kt
```diff
@@ -107,8 +113,10 @@ fun Marker(
     onInfoWindowClick: (Marker) -> Unit = {},
     onInfoWindowClose: (Marker) -> Unit = {},
     onInfoWindowLongClick: (Marker) -> Unit = {},
+    infoWindowComponent: ComposeInfoWindowComponent? = null,
```
adamp, line 116: Why a new type for this instead of `content: @Composable (Marker) -> Unit`?

adamp: This should use the composable function type and not a new type to differentiate overloads on. Use an alternative name for the `Marker` function to differentiate these overloads instead.

app/src/main/java/com/google/maps/android/compose/MapSampleActivity.kt
```diff
@@ -116,12 +122,22 @@ private fun GoogleMapView(modifier: Modifier, onMapLoaded: () -> Unit) {
...
+            Log.d(TAG, "${it.title} was clicked")
+            false
+        }
         Marker(
             position = singapore,
             title = "Zoom in has been tapped $ticker times.",
-            onClick = {
-                println("${it.title} was clicked")
-                false
+            onClick = markerClick
```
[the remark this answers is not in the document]
arriolac: Looks like this is an artifact from reusing the same underlying `ComposeView`. Returning a new `ComposeView` each time appears to address this bug.

README.md
```diff
@@ -95,6 +95,39 @@ GoogleMap(
...
+) {
+    Text(marker.title ?: "Default Marker Title", color = Color.Red)
+}
+```
+
+If you want to customize the entire info window, use the `infoWindow` property
+in the `MapMarker` element instead. Note that if you provide a composable 
+function for `infoWindow`, `content` will be ignored.
+
+```koltin
```
bentrengrove, line 115:
```suggestion
```kotlin
```

README.md
```diff
@@ -95,6 +95,39 @@ GoogleMap(
...
+    //...
+    infoWindow = { marker ->
+        // Implement the custom info window here
+        Column {
+            Text(marker.title ?: "Default Marker Title", color = Color.Red)
+            Text(marker.snippet ?: "Default Marker Snippet", color = Color.Red)
+        }
+    }
+) { marker ->
+    // Don't do this! This will be ignored
```
alexvanyo, line 126: Could `MapMarker` only take `infoWindow`, and not `content` at all?

That would make this issue impossible, where the `content` gets ignored since `infoWindow` is used.

They could both call through to a private `Marker` or `MapMarker` function that takes both to share implementation, but there's no external way to have both be non-null then.

arriolac: That can work and would prevent the issue altogether. Since that would mean the `MapMarker` would only accept a single `@Composable` parameter, `content` should remain and `infoWindow` should be removed to stick with the guideline in [Compose UI layouts](https://github.com/androidx/androidx/blob/androidx-main/compose/docs/compose-api-guidelines.md#compose-ui-layouts).

I would like to rename `MapMarker` to accommodate this change though. Perhaps `MarkerInfoWindow`?

alexvanyo: `MarkerInfoWindow` makes sense! It wouldn't be required, but maybe `Marker` takes no `content`, `MarkerInfoWindow` takes a required `content` for the window, and `MarkerInfoWindowContent` takes a required `content` for the info window content? Not sure if that would be clearer than having the default one have an optional `content`.

arriolac: I like it—way more clear on what `content` is.

bentrengrove (approved):
LGTM!

PR #33, opened 2022-02-15
feat: Add parameter on Marker to control info window visibility state

### LINKED ISSUE #16: Displaying info window as soon as a marker is added

scavazzini:
This pull request adds a new parameter on `Marker` composables to control the info window visibility state.

---

Fixes #16 🦕


arriolac:
Unfortunately the approach of having a state object for the info window's visibility will not work as only one info window can be visible at a time. An info window's visibility is completely managed by the SDK and so I think the best approach is to expose a `show/hide` method on a `MarkerState`. I should have a PR that builds on top of this shortly.

PR #83, opened 2022-03-30
feat: Support specifying animation duration for camera changes.

### LINKED ISSUE #79: Specify animate duration for the map
There is now way to specify the animation duration for the map

I would like to be able to specify the animation duration when calling the `animate` function on the `CameraPositionState` object. Right now I find it too slow for my app.

 **Additional context**
I used the `animateCamera` function from the legacy maps repo before using compose and I may need an option to customize the animation duration.

arriolac:
Fixes #79 🦕

### DIFF
--- a/maps-compose/src/main/java/com/google/maps/android/compose/CameraPositionState.kt
+++ b/maps-compose/src/main/java/com/google/maps/android/compose/CameraPositionState.kt
@@ -33,6 +33,7 @@ import kotlinx.coroutines.Job
 import kotlinx.coroutines.cancel
 import kotlinx.coroutines.currentCoroutineContext
 import kotlinx.coroutines.suspendCancellableCoroutine
+import java.lang.Integer.MAX_VALUE
 import kotlin.coroutines.resume
 import kotlin.coroutines.resumeWithException
 
@@ -175,12 +176,12 @@ class CameraPositionState(
      * This method should only be called from a dispatcher bound to the map's UI thread.
      *
      * @param update the change that should be applied to the camera
-     * @param durationMs The duration of the animation in milliseconds. If null is provided, the
-     * default animation duration will be used. Otherwise, the value provided must be strictly
-     * positive, otherwise an [IllegalArgumentException] will be thrown.
+     * @param durationMs The duration of the animation in milliseconds. If [Int.MAX_VALUE] is
+     * provided, the default animation duration will be used. Otherwise, the value provided must be
+     * strictly positive, otherwise an [IllegalArgumentException] will be thrown.
      */
     @UiThread
-    suspend fun animate(update: CameraUpdate, durationMs: Int? = null) {
+    suspend fun animate(update: CameraUpdate, durationMs: Int = MAX_VALUE) {
         val myJob = currentCoroutineContext()[Job]
         try {
             suspendCancellableCoroutine<Unit> { continuation ->
--- a/maps-compose/src/main/java/com/google/maps/android/compose/CameraPositionState.kt
+++ b/maps-compose/src/main/java/com/google/maps/android/compose/CameraPositionState.kt
@@ -175,11 +175,12 @@ class CameraPositionState(
      * This method should only be called from a dispatcher bound to the map's UI thread.
      *
      * @param update the change that should be applied to the camera
-     * @param durationMillis the duration of the animation in milliseconds. The default animation
-     * duration is used if a negative value is provided.
+     * @param durationMs The duration of the animation in milliseconds. If null is provided, the
+     * default animation duration will be used. Otherwise, the value provided must be strictly
+     * positive, otherwise an [IllegalArgumentException] will be thrown.
      */
     @UiThread
-    suspend fun animate(update: CameraUpdate, durationMillis: Int = -1) {
+    suspend fun animate(update: CameraUpdate, durationMs: Int? = null) {
         val myJob = currentCoroutineContext()[Job]
         try {
             suspendCancellableCoroutine<Unit> { continuation ->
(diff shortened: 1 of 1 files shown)


maps-compose/src/main/java/com/google/maps/android/compose/CameraPositionState.kt
```diff
@@ -173,9 +173,13 @@ class CameraPositionState(
...
      *
      * This method should only be called from a dispatcher bound to the map's UI thread.
+     *
+     * @param update the change that should be applied to the camera
+     * @param durationMillis the duration of the animation in milliseconds. The default animation
+     * duration is used if a negative value is provided.
      */
     @UiThread
-    suspend fun animate(update: CameraUpdate) {
+    suspend fun animate(update: CameraUpdate, durationMillis: Int = -1) {
```
barbeau, line 182: I don't have a strong opinion on this, but the variable name in the underlying Map API is `durationMs`:
https://developers.google.com/android/reference/com/google/android/gms/maps/GoogleMap#animateCamera(com.google.android.gms.maps.CameraUpdate)

However, for the allowed values here, the underlying Maps API throws an `IllegalArgumentException` if this number is not positive:

>durationMs - The duration of the animation in milliseconds. This must be strictly positive, otherwise an IllegalArgumentException will be thrown.

Two questions/comments:
1. Should this API behavior match the underlying Maps SDK, in that `IllegalArgumentException` is thrown if the duration is not valid?
2. I don't believe `0` should be a valid value here based on the above documentation. Seems like if `0` is used, then the `animateCamera()` API shouldn't be called, and the `moveCamera()` API should be used instead.

maps-compose/src/main/java/com/google/maps/android/compose/CameraPositionState.kt
```diff
@@ -235,17 +239,23 @@ class CameraPositionState(
...
             override fun onCancel() {
                 continuation.resumeWithException(CancellationException("Animation cancelled"))
             }
 
             override fun onFinish() {
                 continuation.resume(Unit)
             }
-        })
+        }
+        if (durationMillis < 0) {
```
barbeau, line 254: See note above about `0` not being a valid value.

barbeau (commented):
Thanks @arriolac! Some comments in-line.

maps-compose/src/main/java/com/google/maps/android/compose/CameraPositionState.kt
arriolac: I deliberately did not want `IllegalArgumentException` to be propagated up since the semantics of this method is different (a negative value means the default duration will be used). However, `0` as an input is a good catch in which case calling `moveCamera` would be the right approach. 

On second thought, I'm leaning towards mirroring the SDK's behavior of throwing when `durationMs` is non-positive. We would need to overload `animate` since it won't be possible to set a default duration value without changing the semantics of the API (like I did in this change). This is also more consistent with other design decisions made in this library. Let me know if you have any thoughts here.

maps-compose/src/main/java/com/google/maps/android/compose/CameraPositionState.kt
arriolac: Updated this in [09c990e](https://github.com/googlemaps/android-maps-compose/pull/83/commits/09c990e8b3aaa4f510f236e06aca36770647ca80) to check against null

maps-compose/src/main/java/com/google/maps/android/compose/CameraPositionState.kt
```diff
@@ -173,9 +173,14 @@ class CameraPositionState(
...
      * This method should only be called from a dispatcher bound to the map's UI thread.
+     *
+     * @param update the change that should be applied to the camera
+     * @param durationMs The duration of the animation in milliseconds. If null is provided, the
+     * default animation duration will be used. Otherwise, the value provided must be strictly
+     * positive, otherwise an [IllegalArgumentException] will be thrown.
      */
     @UiThread
-    suspend fun animate(update: CameraUpdate) {
+    suspend fun animate(update: CameraUpdate, durationMs: Int? = null) {
```
barbeau, line 183: I typically prefer values like `MIN_VALUE` or `MAX_VALUE` over `null` when overloading a variable with other semantics (like the "default" behavior here) to keep it from being nullable if nullability isn't otherwise required. Although I don't think I've seen any official best practices for this on Kotlin so other points of view are welcome.

I assume you're not overloading `animate()` to keep the implementation simpler?

arriolac: Yes, ideally a constant value would be nice for `durationMs` but by default the SDK has different durations based on the value of `update`. And yes, I went with this implementation to keep things simple :)

barbeau: Just to clarify, I was thinking of something like this to avoid the nullable `?` and use `MAX_VALUE` as the "default" flag instead of `null`:

```kotlin
    suspend fun animate(update: CameraUpdate, durationMs: Int = MAX_VALUE) {
    ...
        if (durationMs == MAX_VALUE) {
            map.animateCamera(update, cancelableCallback)
        } else {
            map.animateCamera(update, durationMs, cancelableCallback)
        }
```

arriolac: Gotcha—I do like that better. I integrated your feedback in 3f8c1d6

maps-compose/src/main/java/com/google/maps/android/compose/CameraPositionState.kt
```diff
@@ -235,17 +241,23 @@ class CameraPositionState(
     private fun performAnimateCameraLocked(
         map: GoogleMap,
         update: CameraUpdate,
+        durationMs: Int?,
```
barbeau, line 244: This no longer needs to be nullable, right?

arriolac: Yep, the conditional should also check for `MAX_VALUE` 😓  Done in 3f85be9

barbeau (approved):
LGTM!

PR #95, opened 2022-04-14
chore: Use explicitApi.

arriolac:
Use Kotlin's `explicitApi` to encourage explicitly adding a visibility modifier to prevent leaking internal classes.


arriolac:
Closing in favor of #121

PR #97, opened 2022-04-15
chore: Edit maps_compose_dependency region tag.

arriolac:
Add the Maps SDK declaration to the `maps_compose_dependency` region tag and also remove trailing "silent" from `END_EXCLUDE`


sgilsonapigee (approved):
I double checked with https://github.com/googlemaps/android-samples/blob/main/snippets/build.gradle and that looks correct.

PR #137, opened 2022-06-07
fix: Prevent MapView from being recreated when still in composition

### LINKED ISSUE #112: Google Maps Wont restoreState properly when navigating through bottom nav
#### Environment details

1. library: com.google.maps.android:maps-compose:2.1.0

#### Steps to reproduce

  1. clone this [repo](https://github.com/thekingrenz23/GoogleMapsBottomNav)
  2. add your google api key to the manifest file
  3. Quickly navigate between the bottom tabs and check if the maps is loaded properly
  4. When navigating quickly with the bottom tabs the maps wont restore its state eg. camera position, markers

#### Observation
  The maps will load properly even if we navigate quickly on the bottom tabs when the starting route of nav host is the map to reproduce.
   1. Change the AnimatedNavHost start destination from
   ```
AnimatedNavHost(
            navController = navController,
            startDestination = Screen.Other.route
        )
   ```
to
```
AnimatedNavHost(
            navController = navController,
            startDestination = Screen.Map.route
        )
```

and it will work just fine.

Here is a [sample video](https://streamable.com/5afh9a) of the bug

arriolac:
Addresses a bug wherein the underlying `MapView` gets recreated due to lifecycle events, however, the GoogleMap composable never leaves the composition. In this state, map properties (markers, camera position, etc.) never get restored resulting the map centered at (0, 0). The fix is to prevent the map from undergoing reconstruction when it is still in composition (i.e. ON_CREATE is skipped). I tested that this solution works based on the sample provided in #112 as well as the sample in this repo. A bit tricky to create an automated test for this as the test case will need a handle to the raw `GoogleMap` object.

Fixes #112

### DIFF
--- a/maps-compose/src/main/java/com/google/maps/android/compose/GoogleMap.kt
+++ b/maps-compose/src/main/java/com/google/maps/android/compose/GoogleMap.kt
@@ -15,20 +15,18 @@
 package com.google.maps.android.compose
 
 import android.content.ComponentCallbacks
-import android.content.Context
 import android.content.res.Configuration
 import android.location.Location
 import android.os.Bundle
-import android.view.ViewGroup
-import android.widget.FrameLayout
-import android.widget.LinearLayout
 import androidx.compose.foundation.layout.PaddingValues
 import androidx.compose.runtime.Composable
 import androidx.compose.runtime.Composition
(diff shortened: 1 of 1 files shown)


barbeau (approved):
LGTM, thanks @arriolac!

Issue #44, opened 2022-02-21
RFC: Add support for clustering


spoelt:
Do you plan on adding clustering to the Compose version of Google Maps? I tried making the switch from `AndroidView({ MapView() })` to the new Composable function but struggle to implement the `ClusterManager` given that I cannot directly reference the map as an input parameter when instantiating the `ClusterManager`.
(... and thank you for giving us this awesome Composable in the first place 🙏)

jpoehnelt:
@spoelt Thank you for opening this issue. 🙏
Please check out these other resources that might be applicable:

- [Support Console](https://cloud.google.com/support/) - If you have a support contract.
- [Discord](https://discord.gg/9UpV862xZ9) - chat with other developers
- [StackOverflow](http://stackoverflow.com/questions/tagged/google-maps) - use the `google-maps` tag
- [Google Maps Platform issue trackers](https://developers.google.com/maps/support/#issue_tracker)


_This is an automated message, feel free to ignore._

arriolac:
Yes, I would like to add this! Adding clustering support isn't quite that simple though as Maps Compose entirely manages the underlying `GoogleMap`. I've summarized my thoughts on this in [this comment](https://github.com/googlemaps/android-maps-compose/issues/28#issuecomment-1041946666)

StephenVinouze:
Can't this be achieved already with the [Maps Utils Library](https://github.com/googlemaps/android-maps-utils)? Since that's where you get the `ClusterManager`. I haven't tried yet but I naively thought Map Compose would interact with this lib the same way as with Views 🤔

StephenVinouze:
Hmm I see, the `ClusterManager` needs a map as an argument. Never mind my previous comment then 😅

NitroG42:
I think I'll provide some more info on why the cluster manager is needed:
Right now, trying to display a high number of marker on the Maps will make the render freezes for a few seconds
with +500 (+1000) markers, using a simple:
```
poi.forEach {
    Marker(...)
}
```
I think the only solution for now is either display less (based on the viewport for example, restrict to a certain amount or distance), or make an "almost" clustering implementation outside of the composition (based on the viewport, and make "cluster" markers on the fly by grouping your markers based on a distance or lat/long. It might be long to make that of course)

sajithlaldev:
It was a nice feature in Native android. Now I am switched to Jetpack and screwed with my project. :( Hope this will be added soon

barbeau:
[...]

PR #140, opened 2022-06-08
feat: Add experimental MapEffect composable

### LINKED ISSUE #135: Expose raw GoogleMap for extensibility
Currently, there is a need for developers to extend the functionality provided by Maps Compose. For example, there's much need to use features in the Utility Library alongside the Maps Compose library. However, using the Utility Library involves rewriting some of the Maps Compose features manually which is not ideal and brings us back to square one.

Consider an approach that would allow developers to obtain a handle to the `GoogleMap` Maps SDK object to make direct manipulations to it. This would allow developers to extend the map and use utility library features. One approach would be to expose a `MapEffect` composable that exposes the `GoogleMap`

e.g.
```
GoogleMap() {
    MapEffect { googleMap ->
        // use googleMap w/ clustering
    }
}
```

Another option is to provide a CompositionLocal of the `GoogleMap`.

Note: this may have some unintended consequences given that Maps Compose manages the map's properties and so it would have to be made very clear to developers how this should be used.

Proposal in #28 is related.

 **Additional context**
See conversation in #44

arriolac:
Create `MapEffect` composable to expose the underlying `GoogleMap` Maps SDK object to enable extension and usages such as clustering with the utility library.

Example:

```kotlin
@OptIn(MapsComposeExperimentalApi::class)
@Composable
fun GoogleMapClustering(items: List<MyItem>) {
    val cameraPositionState = rememberCameraPositionState {
        position = CameraPosition.fromLatLngZoom(singapore, 10f)
    }
    GoogleMap(
        modifier = Modifier.fillMaxSize(),
        cameraPositionState = cameraPositionState
    ) {
        val context = LocalContext.current
        var clusterManager by remember { mutableStateOf<ClusterManager<MyItem>?>(null) }
        MapEffect(items) { map ->
            clusterManager = ClusterManager<MyItem>(context, map)
            clusterManager?.addItems(items)
        }
        LaunchedEffect(key1 = cameraPositionState.isMoving) {
            if (!cameraPositionState.isMoving) {
                clusterManager?.cluster()
            }
        }
        MarkerInfoWindow(
            state = rememberMarkerState(position = singapore),
            onClick = {
                // This won't work :(
                Log.d(TAG, "I was clicked $it")
                true
            }
        )
    }
}
```

Note: this is an experimental API and requires explicit opt in to `MapsComposeExperimentalApi`.

Fixes #135 🦕

### DIFF
--- a/README.md
+++ b/README.md
@@ -174,6 +174,16 @@ GoogleMap(
         }
         clusterManager?.addItems(items)
     }
+    
+    MarkerInfoWindow(
+        state = rememberMarkerState(position = LatLng(1.35, 103.87)),
+        onClick = {
+            // This won't work :(
+            Log.d("MapEffect", "I cannot be clicked :( $it")
+            true
+        }
+    )
+
 }
 ```
 
@@ -183,9 +193,11 @@ properties while the `GoogleMap` is in composition, and so, setting properties
 on the `GoogleMap` instance provided in the `MapEffect` composable may have
 unintended consequences. For instance, using the utility library to perform 
 clustering as shown in the example above will break `onClick` events from
-being propagated on `Marker` composables. So, if you are using clustering as
-shown above, stick with adding markers through the `ClusterManager` and don't 
-use `Marker` composables (unless you don't care about `onClick` events).
+being propagated on `Marker` composables as shown in the comment above. So, if 
+you are using clustering, stick with adding markers through the `ClusterManager`
+and don't use `Marker` composables (unless you don't care about `onClick` 
+events). Clustering is the only use-case tested with `MapEffect`, there may be
+gotchas depending on what features you use in the utility library.
 
 ## Sample App
 
--- a/README.md
+++ b/README.md
@@ -153,6 +153,40 @@ MarkerInfoWindow(
 }
 ```
 
+#### Obtaining Access to the raw GoogleMap (Experimental)
+
+Certain use cases require extending the `GoogleMap` object to decorate / augment
+the map. For example, while marker clustering is not yet supported by Maps Compose
+(see [Issue #44](https://github.com/googlemaps/android-maps-compose/issues/44)),
+it is desirable to use the available [utility library](https://github.com/googlemaps/android-maps-utils)
+to perform clustering in the interim. Doing so requires access to the Maps SDK
+`GoogleMap` object which you can obtain with the `MapEffect` composable.
+
+```kotlin
+GoogleMap(
+    // ...
+) {
+    val context = LocalContext.current
+    var clusterManager by remember { mutableStateOf<ClusterManager<MyItem>?>(null) }
+    MapEffect(items) { map ->
+        if (clusterManager == null) {
+            clusterManager = ClusterManager<MyItem>(context, map)
+        }
+        clusterManager?.addItems(items)
+    }
+}
+```
+
+Note, however, that `MapEffect` is designed as an escape hatch and has certain
+gotchas. The `GoogleMap` composable provided by the Maps Compose library manages
+properties while the `GoogleMap` is in composition, and so, setting properties
+on the `GoogleMap` instance provided in the `MapEffect` composable may have
+unintended consequences. For instance, using the utility library to perform 
+clustering as shown in the example above will break `onClick` events from
+being propagated on `Marker` composables. So, if you are using clustering as
+shown above, stick with adding markers through the `ClusterManager` and don't 
+use `Marker` composables (unless you don't care about `onClick` events).
+
 ## Sample App
 
 This repository includes a [sample app](app).
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -33,6 +33,7 @@ android {
 
     kotlinOptions {
         jvmTarget = '1.8'
+        freeCompilerArgs += '-Xopt-in=kotlin.RequiresOptIn'
     }
 }
 
@@ -45,6 +46,7 @@ dependencies {
     implementation 'com.google.android.material:material:1.5.0'
     implementation 'com.google.maps.android:maps-ktx:3.3.0'
     implementation "org.jetbrains.kotlin:kotlin-stdlib-jdk7:$kotlin_version"
+    implementation 'com.google.maps.android:android-maps-utils:2.3.0'
 
     androidTestImplementation "androidx.test:core:$androidx_test_version"
     androidTestImplementation "androidx.test:rules:$androidx_test_version"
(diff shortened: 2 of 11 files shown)


app/src/main/java/com/google/maps/android/compose/MapClusteringActivity.kt
```diff
@@ -0,0 +1,98 @@
...
+        var clusterManager by remember { mutableStateOf<ClusterManager<MyItem>?>(null) }
+        MapEffect(items) { map ->
+            if (clusterManager == null) {
+                clusterManager = ClusterManager<MyItem>(context, map)
+            }
+            clusterManager?.addItems(items)
+        }
+        LaunchedEffect(key1 = cameraPositionState.isMoving) {
+            if (!cameraPositionState.isMoving) {
+                clusterManager?.cluster()
```
barbeau, line 71: As written this isn't a 1:1 map to the default behavior of the `ClusterManager` in AMU. The default behavior is driven by `onCameraIdle()`:
https://github.com/googlemaps/android-maps-utils/blob/main/library/src/main/java/com/google/maps/android/clustering/ClusterManager.java#L271

...which then calls through to `cluster()` under certain conditions, some of which are configurable.

I think a better option here would be to call `clusterManager?.onCameraIdle()` to pass the event through?

Or maybe there's a reason not to do this that I'm not aware of.

app/src/main/java/com/google/maps/android/compose/MapClusteringActivity.kt
```diff
@@ -0,0 +1,98 @@
...
+            }
+        }
+        MarkerInfoWindow(
+            state = rememberMarkerState(position = singapore),
+            onClick = {
+                // This won't work :(
+                Log.d(TAG, "I was clicked $it")
```
barbeau, line 78:
```suggestion
                Log.d(TAG, "I cannot be clicked :( $it")
```

README.md
```diff
@@ -153,6 +153,40 @@ MarkerInfoWindow(
...
+
+Note, however, that `MapEffect` is designed as an escape hatch and has certain
+gotchas. The `GoogleMap` composable provided by the Maps Compose library manages
+properties while the `GoogleMap` is in composition, and so, setting properties
+on the `GoogleMap` instance provided in the `MapEffect` composable may have
+unintended consequences. For instance, using the utility library to perform 
+clustering as shown in the example above will break `onClick` events from
+being propagated on `Marker` composables. So, if you are using clustering as
+shown above, stick with adding markers through the `ClusterManager` and don't 
+use `Marker` composables (unless you don't care about `onClick` events).
```
barbeau, line 188: If clustering is the only AMU use case that's been tested (and we provide an example for), should we add a caveat here saying so?

I imagine there could be many ways for this to break with other AMU functionality when mixed with Compose.

barbeau (commented):
Thanks @arriolac! Some comments in-line.

While there is certainly demand for this feature and this seems like the most-likely near-term fix, I also wonder if this is opening a Pandora's box for all kinds of bugs. There are already known gotchas with AMU and using multiple features simultaneously, and it seems like this is piling on top of those. I do like marking this as experimental, largely for this reason. Any other thoughts on mitigating potentially bad experiences is welcome.

ColtonIdle:
For what it's worth, I love the idea of this being added in an experimental state for the time being. Since I can't use clustering in android-maps-compose, using this library is a non-starter, (which is a shame because this library already includes a bunch of fixes and stuff I've run into using AndroidVIew + google maps **without clustering**. see: the lifecycle issue created in the issuetracker when you use AndroidVIew + google maps in a tab bar layout)

It seems like knowing that this is a temporary solution would at least allow way more people to adopt this library, and then in the 6months+ it takes to add clustering as a first class citizen we could at least build up to that.

of course. im bias because i tore my hair our trying to get androidView + google maps + tabs + clustering to work so i definitely rather use something that makes no promises, but is kinda supported by the maintainers here + community, vs the frankenstein thing I hacked together. lol

cheers and thanks for considering

app/build.gradle
```diff
@@ -60,9 +62,9 @@ dependencies {
     // module.
     // However, this should remain uncommented on the `main` branch so that
     // the maven declaration of Maps Compose can be used as a snippet.
-    // implementation project(':maps-compose')
+    implementation project(':maps-compose')
```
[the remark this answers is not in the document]
arriolac: Yes, I think this should be reverted. I wanted to keep the PR open with this to validate that checks passed. Note that when I revert this checks will break so we'll have to merge with failed checks and then update the sample app's version once the new `2.3.0` is available.

app/src/main/java/com/google/maps/android/compose/MapClusteringActivity.kt
arriolac: Oh, yeah good call. This code should call `clusterManager?.onCameraIdle()` instead.

arriolac:
Thanks @ColtonIdle for providing color to your use case. I thought of this feature precisely to address the issue you faced—it seems to be a really common workflow based on reports I've seen from the community.

> Thanks @arriolac! Some comments in-line.
> 
> While there is certainly demand for this feature and this seems like the most likely near-term fix, I also wonder if this is opening a Pandora's box for all kinds of bugs. There are already known gotchas with AMU and using multiple features simultaneously, and it seems like this is piling on top of those. I do like marking this as experimental, largely for this reason. Any other thoughts on mitigating potentially bad experiences is welcome.

@barbeau this is definitely not a feature that I originally intended on adding (I think we had a convo about why we choose not to expose the raw `GoogleMap` before). However, looking at some of the common themes the community is facing (as @ColtonIdle has pointed out), an escape hatch is very much needed for extensibility. My 2c is that as long as we have the right guardrails in place (docs, explicit opt-in with the new `MapsComposeExperimentalApi`), this solution is much better than developers having to copy/paste parts of Maps Compose code in their projects to get a feature like clustering to work. This is also considering that a Compose version of clustering is yet to be designed and implemented which is no easy feat. Lmk if this makes sense—I'm open to more ideas on how we can integrate this safely for the community.

barbeau (approved):
Thanks @arriolac. I agree that all things considered the best way to move forward at this point is with this as an experimental feature.

newmanw:
This is awesome, thanks @arriolac!  

Curious about a couple things, not that I am looking for anything specific, just curious.
1. Are there still plans to implement maps compose clustering?
2. Currently maps compose is just a wrapper around the maps sdk, will this ever go "compose first"?

arriolac:
> This is awesome, thanks @arriolac!
> 
> Curious about a couple things, not that I am looking for anything specific, just curious.
> 
> 1. Are there still plans to implement maps compose clustering?
> 2. Currently maps compose is just a wrapper around the maps sdk, will this ever go "compose first"?

Hey @newmanw, thanks for the questions.

1. This is definitely still desirable. I imagine once that's completed `MapEffect` should be evaluated for deprecation unless there's another common use case folks want to use it for. You can follow #44.
2. Android is pretty invested in Compose so I imagine this _could_ happen. No plans atm as far as I'm aware.

PR #148, opened 2022-06-15
chore: Have app depend on local library vs. the published Maven Central artifact

barbeau:
The project is currently deadlocked in a failing CI state as a result of the app depending on the latest Maven Central release of the library. We can't update the app to the newest release from Maven Central so CI passes because the release doesn't exist yet, and we can't publish a new release because the CI is failing:
https://github.com/googlemaps/android-maps-compose/runs/6889770989?check_suite_focus=true

This PR changes the app to depend on the local library so CI passes and the new release of the library can be published following the merge of this PR. Then we have the choice of switching the app back to depending on Maven Central release, or leaving depending on the local library.

@arriolac @wangela What are your thoughts on leaving the app depending on the local library? We have to keep switching back and forth, and the process includes merging failed CI builds by design, because we have it pinned to the last Maven Central release. Seems less risky to me to leave it depending on the local library so we have consistently passing CI.

---

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -37,9 +37,7 @@ android {
     }
 }
 
-// [START maps_android_compose_dependency]
 dependencies {
-    // [START_EXCLUDE silent]
     implementation 'androidx.activity:activity-compose:1.4.0'
     implementation "androidx.compose.foundation:foundation:$compose_version"
     implementation "androidx.compose.material:material:$compose_version"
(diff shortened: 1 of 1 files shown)


arriolac:
The main issue integrating this change is that it breaks the region tag `maps_android_compose_dependency` which is used to display the latest version in https://developers.google.com/maps/documentation/android-sdk/maps-compose#installation

This change is ok though the region tag would have to be migrated elsewhere (perhaps in https://github.com/googlemaps/android-samples)

barbeau:
>The main issue integrating this change is that it breaks the region tag maps_android_compose_dependency which is used to display the latest version in https://developers.google.com/maps/documentation/android-sdk/maps-compose#installation
>
>This change is ok though the region tag would have to be migrated elsewhere (perhaps in https://github.com/googlemaps/android-samples)

Ah, thanks, I forgot about that. Yes, I suppose the `// [START maps_android_compose_dependency]` region tag could live as a dependency in one of the android-samples apps that doesn't actually include any maps-compose functionality? This way it would always get bumped by dependabot but the build wouldn't break because of backwards/forwards incompatible changes.

arriolac:
FYI @wangela to coordinate the above doc change described above

barbeau:
I've opened a PR on the android-samples repo which I believe should migrate the region tags as discussed above:
https://github.com/googlemaps/android-samples/pull/995

wangela (approved):
Thank you, good fix for the sample app.

PR #150, opened 2022-06-16
Rough proposed new API surface for marker dragging

### LINKED ISSUE #149: Marker dragging: fix loss of drag events and difficulty observing marker positions
android-maps-compose 2.2.1

The current API design of `Marker` and `MarkerState` poses significant challenges. This issue is a continuation of comments I made in #10; @arriolac asked to move discussion to a new issue.

1. 'DragState' events are lost, in particular `START`, but also `DRAG`, and, theoretically, even `END`.

This is due to the characteristics of snapshot state in relation to events:

> Observable state is a lossy compression of the events that produced that state.

https://developer.android.com/reference/kotlin/androidx/compose/runtime/package-summary#snapshotFlow(kotlin.Function0)

I have a use case where I need to reliably detect `START` and `END` to temporarily keep a backup of the old marker location to restore in case the final marker drag location is deemed invalid by business logic.

2. [Hoisted state objects](https://github.com/androidx/androidx/blob/androidx-main/compose/docs/compose-api-guidelines.md#hoisted-state-types) are not suitable for representing collections of app marker data.

Hoisted state objects can be a suitable vessel for maintaining UI state. However, they are an anti-pattern for representing data that is ultimately maintained in an app data model. Observing requires using `snapshotFlow` for each indivdual state object, and updating implies feeding model data back to the state object in some manner.

An app commonly maintains a collection of markers, so a collection of `MarkerState` objects is needed, one for each marker. Markers may be added or removed, adding significant complexity to the task of keeping separate collections of marker data in sync. Typically one will opt to recreate the entire collection if one element changes in any way.

I will add more detail and proposals in comments below.

Clarification: In my main use case for this feature, markers define the corners of a polyline/polygon, so order is significant, and the marker collection is a list, where updates, inserts, and deletions can occur at any position. Keeping lists of variable size with changing marker content synchronized is a more challenging problem than for example maps of markers where order is not significant; I use the latter in another case, and it is easier to deal with.

bubenheimer:
Fixes #149  🦕

This is primarily intended at this point to share the changed API surface I have in mind. Underlying rationale is in linked issue.

### DIFF
--- a/maps-compose/src/main/java/com/google/maps/android/compose/Marker.kt
+++ b/maps-compose/src/main/java/com/google/maps/android/compose/Marker.kt
@@ -14,17 +14,7 @@
 
 package com.google.maps.android.compose
 
-import androidx.compose.runtime.Composable
-import androidx.compose.runtime.ComposeNode
-import androidx.compose.runtime.CompositionContext
-import androidx.compose.runtime.Immutable
-import androidx.compose.runtime.currentComposer
-import androidx.compose.runtime.getValue
-import androidx.compose.runtime.mutableStateOf
-import androidx.compose.runtime.rememberCompositionContext
-import androidx.compose.runtime.saveable.Saver
-import androidx.compose.runtime.saveable.rememberSaveable
-import androidx.compose.runtime.setValue
+import androidx.compose.runtime.*
 import androidx.compose.ui.geometry.Offset
 import com.google.android.gms.maps.model.BitmapDescriptor
 import com.google.android.gms.maps.model.LatLng
@@ -62,22 +52,13 @@ public enum class DragState {
 }
 
 /**
- * A state object that can be hoisted to control and observe the marker state.
- *
- * @param position the initial marker position
+ * A state object that can be hoisted to observe the marker state.
  */
-public class MarkerState(
-    position: LatLng = LatLng(0.0, 0.0)
-) {
+public class MarkerState {
     /**
-     * Current position of the marker.
+     * Current dragging state of the marker.
      */
-    public var position: LatLng by mutableStateOf(position)
-
-    /**
-     * Current [DragState] of the marker.
-     */
-    public var dragState: DragState by mutableStateOf(DragState.END)
+    public var dragging: Boolean by mutableStateOf(false)
         internal set
 
     // The marker associated with this MarkerState.
@@ -103,25 +84,10 @@ public class MarkerState(
     public fun hideInfoWindow() {
         marker?.hideInfoWindow()
     }
-
-    public companion object {
-        /**
-         * The default saver implementation for [MarkerState]
-         */
-        public val Saver: Saver<MarkerState, LatLng> = Saver(
-            save = { it.position },
-            restore = { MarkerState(it) }
-        )
-    }
 }
 
 @Composable
-public fun rememberMarkerState(
-    key: String? = null,
-    position: LatLng = LatLng(0.0, 0.0)
-): MarkerState = rememberSaveable(key = key, saver = MarkerState.Saver) {
-    MarkerState(position)
-}
+internal fun rememberMarkerState(): MarkerState = remember { MarkerState() }
 
 /**
  * A composable for a marker on the map.
@@ -148,10 +114,10 @@ public fun rememberMarkerState(
 @Composable
 @GoogleMapComposable
 public fun Marker(
+    position: LatLng,
     state: MarkerState = rememberMarkerState(),
     alpha: Float = 1.0f,
     anchor: Offset = Offset(0.5f, 1.0f),
-    draggable: Boolean = false,
     flat: Boolean = false,
     icon: BitmapDescriptor? = null,
     infoWindowAnchor: Offset = Offset(0.5f, 0.0f),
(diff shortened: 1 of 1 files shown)


maps-compose/src/main/java/com/google/maps/android/compose/Marker.kt
```diff
@@ -148,10 +114,10 @@ public fun rememberMarkerState(
 @Composable
 @GoogleMapComposable
 public fun Marker(
+    position: LatLng,
     state: MarkerState = rememberMarkerState(),
     alpha: Float = 1.0f,
     anchor: Offset = Offset(0.5f, 1.0f),
-    draggable: Boolean = false,
```
arriolac, line 154: What's the reason for removing this param? I would expect this to still be here.

maps-compose/src/main/java/com/google/maps/android/compose/Marker.kt
```diff
@@ -148,10 +114,10 @@ public fun rememberMarkerState(
 @Composable
 @GoogleMapComposable
 public fun Marker(
+    position: LatLng,
```
arriolac, line 117: This is definitely the Compose way of representing a marker's position. However, since this is an interop library, the issue with this design is that the marker's position is owned by the Maps SDK. So when you drag the marker to a new location, this value is no longer correct. This is why `MarkerState` was introduced and needed as a workaround to the internal behavior of the Maps SDK.

arriolac (commented):
Some questions on your rough design. I'm hoping the addition of an event lambda specific for drag events will resolve the main issue you're encountering with the current design.

I would expect a design closer to this:

```diff
@Composable
fun Marker(
  // ...
+ onDragEvent: ((DragState) -> Unit)? = null,
)

public class MarkerState(
   position: LatLng = LatLng(0.0, 0.0)
) {
   public var position: LatLng by mutableStateOf(position)

- public var dragState: DragState by mutableStateOf(DragState.END)
+ public var dragging: Boolean by mutableStateOf(false)
}
```

The reason the position still has to be encapsulated within `MarkerState` is because the Maps SDK is the owner of the marker's position, which is not the Compose way of doing this. This solution is a workaround that behavior (this is similar to why we have `CameraPositionState`). Let me know if this makes sense and if I'm missing something.

bubenheimer:
Thank you for the input, @arriolac I'm OOO right now, but should have some
time to get back into this in 2 weeks

maps-compose/src/main/java/com/google/maps/android/compose/Marker.kt
javadude: I was thinking the same thing - we still need it to indicate if the user _can_ drag a marker

maps-compose/src/main/java/com/google/maps/android/compose/Marker.kt
```diff
@@ -161,16 +127,17 @@ public fun Marker(
     title: String? = null,
     visible: Boolean = true,
     zIndex: Float = 0.0f,
+    onDragEvent: ((DragState, LatLng) -> Unit)? = null,
```
javadude, line 130: maybe just onDrag(...)?

bubenheimer: Not really opposed to it, except for English semantics, which are debatable. To me `onDrag` sounds like a singular complete drag action to a new position. Marker dragging is a sequence of events, at least 2, usually many more. `onDragEvent` carries a less succinct meaning to me, it could mean anything. `onDragSequenceEvent` would be better from that perspective, but is verbose.

maps-compose/src/main/java/com/google/maps/android/compose/Marker.kt
bubenheimer: Thanks, having `DragState` available via an event handler will be useful.

In #149 I laid out in detail why I find it both necessary and feasible to also eliminate `position` from `MarkerState`. In summary, it is true that the Maps SDK owns the marker's position, but closely aligning the Compose API design with that ownership model is barely usable except in toy apps. I linked several issues caused by this design that came up since I created #149, and I am certain that this pattern will repeat until the design changes.

I agree that it is not ideal and means taking a small risk to turn the ownership model on its head, but the current approach is just too awful to use, every time I look at my workarounds there I get a stomach ache. A design change here could be confined to an experimental playground somehow.

I'd likely try to run my own fork if it cannot be resolved within the main project.

`CameraPositionState` is very different because there is only one camera. Hoisting it into a State object does not have a great impact on usability.

maps-compose/src/main/java/com/google/maps/android/compose/Marker.kt
bubenheimer: Please see my comments in #149 regarding how to keep marker position up-to-date.

maps-compose/src/main/java/com/google/maps/android/compose/Marker.kt
bubenheimer: I do think that `onDragEvent` should receive the current Marker position as a parameter, as I suggested in my original PR. Otherwise the `onDragEvent` lambda would need to rely on an implicit promise that MarkerState is timely updated to the position tied to the corresponding drag event. This is obscure on the API consumer's side, and would seem error-prone on the API implementation side.

bubenheimer:
After thinking about it some more I will take another shot at reworking my code under the assumption of MarkerState continuing to encapsulate the marker position. I expect to have an update in a few days.

maps-compose/src/main/java/com/google/maps/android/compose/Marker.kt
arriolac: Moving `position` here would definitely match the mental model of Compose though I think changing the behavior of what the Maps SDK is doing under the hood to create a better Compose API might create more problems down the line. 

I see your comment below about reworking a solution. My 2c of a summary of proposed changes that would solve your particular use case is:
* Add `onDrag: (LatLng, DragEvent) ->` lambda to `Marker`
* Deprecate `MarkerState.dragState`
* Add `MarkerState.isDragging`

Issue #49, opened 2022-02-28
Source of camera movement -  Exposing GoogleMap.OnCameraMoveStartedListener "reason"


webfrea-k:
Hi!

Is there an equivalent for `OnCameraMoveStartedListener.REASON_GESTURE`? If not, can you please support it? I would like to trigger a specific action when users moves the map. I've searched `CameraPositionState.kt` and found `isMoving` but this works the same if the camera was moved by code or by user.

jpoehnelt:
@webfrea-k Please take a moment to fill out [this short survey](https://forms.gle/wYuzN6UibnFj7ZL29). Thank you!

_This is an automated message, feel free to ignore._

polivmi1:
We do have the same use case. Some location with markers is being followed and when the user moves the map, the automatic "following" will stop. GoogleMap.OnCameraMoveStartedListener.REASON_GESTURE will tell us exactly that.

polivmi1:
As a workaround I could use a detection modifier which doesn't consume the events:
`.pointerInput(Unit) {
                forEachGesture {
                    awaitPointerEventScope {
                        awaitFirstDown(requireUnconsumed = false)
                        actionOnReasonGesture()
                    }
                }
            },`

arriolac:
This can be supported as another property on `CameraPositionState`

PR #154, opened 2022-06-22
feat: Add the reason for map camera movement

barbeau:
As discussed in https://github.com/googlemaps/android-maps-compose/issues/49, this PR exposes the `reason` value passed back in `GoogleMap.OnCameraMoveStartedListener(int reason)` callback so that developers can tell the reason that the camera started moving (e.g., user gesture, developer animation, API animation):
https://developers.google.com/android/reference/com/google/android/gms/maps/GoogleMap.OnCameraMoveStartedListener#onCameraMoveStarted(int)

For constant values see:
https://developers.google.com/android/reference/com/google/android/gms/maps/GoogleMap.OnCameraMoveStartedListener#constants

The PR also adds demo code to `LocationTrackingActivity` to print the reason that the camera moves, as well as assertions to test behavior in an existing unit test.

---

Fixes https://github.com/googlemaps/android-maps-compose/issues/49 🦕

### DIFF
--- a/maps-compose/src/main/java/com/google/maps/android/compose/CameraPositionState.kt
+++ b/maps-compose/src/main/java/com/google/maps/android/compose/CameraPositionState.kt
@@ -27,12 +27,7 @@ import com.google.android.gms.maps.GoogleMap
 import com.google.android.gms.maps.Projection
 import com.google.android.gms.maps.model.CameraPosition
 import com.google.android.gms.maps.model.LatLng
-import kotlinx.coroutines.CancellableContinuation
-import kotlinx.coroutines.CancellationException
-import kotlinx.coroutines.Job
-import kotlinx.coroutines.cancel
-import kotlinx.coroutines.currentCoroutineContext
-import kotlinx.coroutines.suspendCancellableCoroutine
+import kotlinx.coroutines.*
 import java.lang.Integer.MAX_VALUE
 import kotlin.coroutines.resume
 import kotlin.coroutines.resumeWithException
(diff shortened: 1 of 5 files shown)


arriolac (commented):
Change generally LGTM although I think this is a good opportunity to improve the type safety around `reason` and use an enum instead. A new type can also be created for the initial case. Note that this is a pattern already used in the library for `MapType`.

barbeau:
@arriolac Good idea! I convert to an enum in https://github.com/googlemaps/android-maps-compose/pull/154/commits/819179ec6d98e27b54740a8b6c0eefa5ae4dcee3.

I wasn't sure how to handle the case where a new Maps SDK constant is passed back from the `GoogleMap.OnCameraMoveStartedListener` that isn't known to the maps-compose library (e.g., if the Maps SDK is updated). I ended up mapping these unknown constants to another `CameraMoveStartedReason` enum value ` UNKNOWN(-2)`, but let me know if you have a better idea of how to handle this.

maps-compose/src/main/java/com/google/maps/android/compose/CameraPositionState.kt
```diff
@@ -27,12 +27,7 @@ import com.google.android.gms.maps.GoogleMap
 import com.google.android.gms.maps.Projection
 import com.google.android.gms.maps.model.CameraPosition
 import com.google.android.gms.maps.model.LatLng
-import kotlinx.coroutines.CancellableContinuation
-import kotlinx.coroutines.CancellationException
-import kotlinx.coroutines.Job
-import kotlinx.coroutines.cancel
-import kotlinx.coroutines.currentCoroutineContext
-import kotlinx.coroutines.suspendCancellableCoroutine
+import kotlinx.coroutines.*
```
arriolac, line 30: Nit: Use single name imports here. I have my preferences configured to the following:

image

This should probably be enforced in the linter.

barbeau: Ah, good catch - I fixed the imports to not use wildcards and updated my AS settings.

PR #168, opened 2022-07-01
fix: DisappearingScaleBar do not disappear after specified visibilityDurationMillis

### LINKED ISSUE #166: DisappearingScaleBar do not disappear after specified visibilityDurationMillis
The new DisappearingScaleBar does not disappear after the specified visibilityDurationMillis. Unless I have interpreted it wrong, I think what DisappearingScaleBar does is that it automatically disappears after specified visibilityDurationMillis on every zoom click.

But currently, after the initial disappearance that is not happening.

#### Steps to reproduce

  1. Add DisappearingScaleBar to map
  2. Wait for it to disappear for the first time
  3. Just press zoom in one time
  4. After the zoom in the scalebar does not automatically disappear.

I have also attached a video for the issue. It only contains the DisappearingScaleBar on top of the map.

https://user-images.githubusercontent.com/20212314/176884199-d9bc0e42-d426-453a-aaac-b967a6c3e2bc.mp4

. 

#### Code example
```
GoogleMap(
  modifier = Modifier.matchParentSize(),
  onMapLoaded = {
    isMapLoaded = true
  }
)
DisappearingScaleBar(
  modifier = Modifier
  .padding(top = 5.dp, end = 15.dp)
  .align(Alignment.TopStart),
  visibilityDurationMillis = 5_000
)
```

I think the issue lies in the LaunchedEffect of  DisappearingScaleBar. Because of the condition, it is not able to hide the scale bar after the time-out period.

Priyankkjain:
Addresses a fix for #166 where DisappearingScaleBar does not disappear after specified visibilityDurationMillis. I think it's good to remove the condition so that the visibility and delay will restart every time the zoom changes. This will make sure that after given visibilityDurationMillis the ScaleBar will disappear. Also, any zoom change will restart the LaunchedEffect again so that the old time-out will not be an issue.

I would love to hear the feedback if this is the correct way to address the issue or not. Any comments will be valuable.

Fixes #166

### DIFF
--- a/maps-compose-widgets/src/main/java/com/google/maps/android/compose/widgets/ScaleBar.kt
+++ b/maps-compose-widgets/src/main/java/com/google/maps/android/compose/widgets/ScaleBar.kt
@@ -213,14 +213,11 @@ public fun DisappearingScaleBar(
     }
 
     LaunchedEffect(key1 = cameraPositionState.position.zoom) {
-        if (visible.isIdle && !visible.currentState) {
-            // Show ScaleBar
-            visible.targetState = true
-        } else if (visible.isIdle && visible.currentState) {
-            // Hide ScaleBar after timeout period
-            delay(visibilityDurationMillis.toLong())
-            visible.targetState = false
-        }
+        // Show ScaleBar
+        visible.targetState = true
+        delay(visibilityDurationMillis.toLong())
+        // Hide ScaleBar after timeout period
+        visible.targetState = false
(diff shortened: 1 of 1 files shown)


barbeau (approved):
Thanks @Priyankkjain, LGTM. Seems like the existing implementation is getting hung up on the `visible.isIdle` being `false` and not triggering the hide animation, so this seems like a good fix.

Issue #174, opened 2022-07-12
Fix flaky tests on CI


arriolac:
#### Steps to reproduce

Run the test workflow on `main` and notice that instrumentation tests are failing. Example failed runs:

https://github.com/googlemaps/android-maps-compose/runs/7306842111
https://github.com/googlemaps/android-maps-compose/runs/7283340731

Also see discussion in #171

#### Code example

N/A

#### Stack trace

```
com.google.maps.android.compose.MapInColumnTests > testScrollColumn_MapCameraRemainsSame[test(AVD) - 10] FAILED 
	java.lang.AssertionError: Assert failed: The component is displayed!
	at androidx.compose.ui.test.AssertionsKt.assertIsNotDisplayed(Assertions.kt:51)
```

arriolac:
Looks like there are 2 causes for the failed actions linked in the description:

1. Failure due to secrets not being available on forks. This is the reason for the failed PR in https://github.com/googlemaps/android-maps-compose/runs/7283340731. This should be addressed by #175 
2. Failure in `MapInColumnTests.testScrollColumn_MapCameraRemainsSame`. This appears to be env specific as running it locally passes. Perhaps running the test on the same emulator might give us some hints on what might be causing the test failure.

barbeau:
I tried running locally on this emulator and the tests still pass:

![image](https://user-images.githubusercontent.com/928045/178588913-65601e4e-29b6-4891-b8b9-94bf0776c615.png)

wangela:
Tests cannot pass on PRs from forks. Root cause: https://github.com/community/community/discussions/50161

PR #175, opened 2022-07-12
chore: Use pull_request_target.

arriolac:
Use `pull_request_target` so that PRs from forked repos have access to secrets. This should fix failed runs due to the Maps API key not being injected: https://github.com/googlemaps/android-maps-compose/runs/7283340731

Relates to #174  🦕

### DIFF
--- a/.github/workflows/instrumentation-test.yml
+++ b/.github/workflows/instrumentation-test.yml
@@ -22,6 +22,8 @@ on:
     branches-ignore: ['gh-pages']
   pull_request:
     branches-ignore: ['gh-pages']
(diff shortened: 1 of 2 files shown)


barbeau:
@arriolac Doesn't look like the tests ran?

image

arriolac:
Hmmm maybe we need both `pull_request` and `pull_request_target`. Will update

arriolac:
Looks like instrumentation tests are failing since this PR is coming from a fork. Will reopen it from a branch in this repo and then PRs from forks should start working.

PR #176, opened 2022-07-12
chore: Use pull_request_target.

arriolac:
Use pull_request_target so that PRs from forked repos have access to secrets. This should fix failed runs due to the Maps API key not being injected: https://github.com/googlemaps/android-maps-compose/runs/7283340731

Relates to https://github.com/googlemaps/android-maps-compose/issues/174 🦕

Same as #175 but opening this PR from a branch.


barbeau (approved):
Looks like CI failure is same as previously discussed with MapInColumnTests, so let's merge this

wangela (approved):

PR #179, opened 2022-07-14
fix: GoogleMap breaking preview mode.

### LINKED ISSUE #128: Preview for GoogleMap Composable
Screenshot 2022-05-25 235552
I thought there was an issue with my Android Studio, but when I try to render a preview for my UI that has `GoogleMap` composable, it literally shows only a blank preview. It is so difficult working on a UI that has `GoogleMap `composable in it, since I have to build my app to the emulator to see how my UI looks like with the map. 

It'd be helpful if we could see a preview of our map view even if it is only a static map.

arriolac:
Fix preview mode by checking if the GoogleMap composable was invoked within a preview. This allows previews to be shown
for other UI elements, but, the map will still not be rendered.

Fixes #128 🦕

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -45,6 +45,8 @@ dependencies {
     implementation 'com.google.maps.android:maps-ktx:3.3.0'
     implementation "org.jetbrains.kotlin:kotlin-stdlib-jdk7:$kotlin_version"
     implementation 'com.google.maps.android:android-maps-utils:2.3.0'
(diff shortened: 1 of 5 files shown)


barbeau (approved):
Thanks @arriolac! LGTM. Tests pass locally for me, and it looks like the CI is failing with the same flaky test discussed in PR https://github.com/googlemaps/android-maps-compose/pull/180, so we can merge.

PR #188, opened 2022-08-14
fix: Prevent a memory leak on AndroidView which may not call ON_DESTROY

### LINKED ISSUE #138: Memory leak
I have mentioned this in https://github.com/googlemaps/android-maps-compose/issues/26 , but it is already closed, so I am opening a new issue.
I am getting a leak if the map component is cleared and then the activity is destroyed.

Previously when implementing it through AndroidView, to get rid of the maps leak, I had to use:

```
DisposableEffect(lifecycle) {
        lifecycle.addObserver(lifecycleObserver)
        onDispose {
            lifecycle.removeObserver(lifecycleObserver)
            mapView.onDestroy()
            mapView.removeAllViews()
        }
    }
```

And 

```
remember(mapView) {
        LifecycleEventObserver { _, event ->
            when (event) {
                Lifecycle.Event.ON_CREATE -> mapView.onCreate(Bundle())
                Lifecycle.Event.ON_START -> mapView.onStart()
                Lifecycle.Event.ON_RESUME -> mapView.onResume()
                Lifecycle.Event.ON_PAUSE -> mapView.onPause()
                Lifecycle.Event.ON_STOP -> mapView.onStop()
                Lifecycle.Event.ON_DESTROY -> {
                    //handled in onDispose
                }
                else -> throw IllegalStateException()
            }
        }
    }
```

The reason is that Lifecycle.Event.ON_DESTROY might not be called, and this way you would always dispose of the map view in onDispose, which will always be called.
Would you be able to add this? I did fork and update it and the big memory leak disappeared.

`1 APPLICATION LEAKS
    
    References underlined with "~~~" are likely causes.
    Learn more at https://squ.re/leaks.
    
    56446 bytes retained by leaking objects
    Signature: 34a030bd011ab1ccd655bbeac3fbc8465c53ce5f
    ┬───
    │ GC Root: Thread object
    │
    ├─ com.google.maps.api.android.lib6.gmm6.vector.n instance
    │    Leaking: UNKNOWN
    │    Retaining 3.5 MB in 26933 objects
    │    Thread name: 'RenderDrive'
    │    ↓ n.e
    │        ~
    ├─ com.google.maps.api.android.lib6.gmm6.vector.p instance
    │    Leaking: UNKNOWN
    │    Retaining 3.5 MB in 26930 objects
    │    ↓ p.f
    │        ~
    ├─ com.google.maps.api.android.lib6.gmm6.api.ac instance
    │    Leaking: UNKNOWN
    │    Retaining 3.5 MB in 26929 objects
    │    View not part of a window view hierarchy
    │    View.mAttachInfo is null (view detached)
    │    View.mWindowAttachCount = 1
    │    mContext instance of ExampleApplication
    │    ↓ ac.N
    │         ~
    ├─ com.google.maps.api.android.lib6.impl.ax instance
    │    Leaking: UNKNOWN
    │    Retaining 20 B in 1 objects
    │    a instance of ExampleApplication
    │    b instance of ExampleActivity with mDestroyed = true
    │    ↓ ax.b
    │         ~
    ╰→ ExampleActivity instance`

polivmi1:
This PR fixes a memory leak that can happen in AndroidView when navigating to another screen.
The Lifecycle ON_DESTROY event is never called because in onDispose we remove the mapLifecycleObserver and this method is called before the Lifecycle ON_DESTROY event.

We have to handle any Lifecycle ON_DESTROY in the onDispose method.

---

Fixes #138 🦕

### DIFF
--- a/maps-compose/src/main/java/com/google/maps/android/compose/GoogleMap.kt
+++ b/maps-compose/src/main/java/com/google/maps/android/compose/GoogleMap.kt
@@ -175,6 +175,8 @@ private fun MapLifecycle(mapView: MapView) {
         onDispose {
             lifecycle.removeObserver(mapLifecycleObserver)
             context.unregisterComponentCallbacks(callbacks)
+            mapView.onDestroy()
+            mapView.removeAllViews()
         }
     }
 }
@@ -195,7 +197,9 @@ private fun MapView.lifecycleObserver(previousState: MutableState<Lifecycle.Even
             Lifecycle.Event.ON_RESUME -> this.onResume()
             Lifecycle.Event.ON_PAUSE -> this.onPause()
             Lifecycle.Event.ON_STOP -> this.onStop()
-            Lifecycle.Event.ON_DESTROY -> this.onDestroy()
+            Lifecycle.Event.ON_DESTROY -> {
+                //handled in onDispose
+            }
             else -> throw IllegalStateException()
         }
         previousState.value = event


polivmi1:
I signed it.

maps-compose/src/main/java/com/google/maps/android/compose/GoogleMap.kt
```diff
@@ -175,6 +175,8 @@ private fun MapLifecycle(mapView: MapView) {
         onDispose {
             lifecycle.removeObserver(mapLifecycleObserver)
             context.unregisterComponentCallbacks(callbacks)
+            mapView.onDestroy()
+            mapView.removeAllViews()
```
arriolac, line 179: Is this line necessary? I would think `mapView.onDestroy()` should suffice.

polivmi1: I was encountering a google map leak similar to this one: https://stackoverflow.com/questions/65358900/leakcanary-detecting-a-leak-in-my-java-android-app-google-maps and only by removing the views did it get cleaned by GC. It happened when having a route with markers. Do you want me to reproduce locally and add screenshots + leakcanary logs? This is potentially something that could be fixed in google library. There developers can override onDestroy and fix the leak. In this wrapper, they can't.

arriolac: Thanks for clarifying. Let's keep this line here as I don't see any impact as the MapView is being removed/destroyed anyway. I'd like to merge in #193 before merging this in.

arriolac (approved):

polivmi1:
@arriolac I see the test problems have been fixed - can we re-run this and merge unless there is no problem?

arriolac:
This should be good to merge. Will let @wangela take over from here

wangela (approved):

wangela:
Thank you for your contribution @polivmi1 !

wangela:
So sorry @polivmi1, would you mind `git commit --amend` the commit message in your branch to:

```
fix: Prevent a memory leak on AndroidView which may not call ON_DESTROY
```

and then push the branch again? The reason for this is the original commit message did not follow conventional commit conventions, and was interpreted by our semantic release bot as not triggering a new release when in fact it is a bugfix and should trigger a patch release. I have reverted the merge from this PR so you can re-submit your changes.

Once you push the branch, it will create a new PR and you can reference this PR #188 and the original issue that is fixed #138 in the new PR's description, to link to the historical discussion preserved in this PR review.

Apologies for not catching the commit message format during the main review.

polivmi1:
@wangela I have changed the commit message in https://github.com/polivmi1/android-maps-compose/tree/138_memory_leak it hasn't created a new PR (this is already in state merged). Should I create a completely new PR and mention this one there?

wangela:
@polivmi1 Yes, I see a "Create pull request" button here that you can use: https://github.com/googlemaps/android-maps-compose/compare/main...polivmi1:android-maps-compose:138_memory_leak

thekingrenz23:
Waiting for this release to be fixed

polivmi1:
@wangela see https://github.com/googlemaps/android-maps-compose/pull/206

PR #193, opened 2022-08-18
chore: Disable flaky test.

arriolac:
Disables flaky test on CI.

I decided to disable it since commits have been landing to `main` despite the test failing. It's a known env-specific issue and  I think disabling this is the right move so that merges aren't accidentally made when a different failure happens. This should be uncommented again and addressed as part of https://github.com/googlemaps/android-maps-compose/issues/174.

### DIFF
--- a/.github/workflows/instrumentation-test.yml
+++ b/.github/workflows/instrumentation-test.yml
@@ -22,8 +22,6 @@ on:
     branches-ignore: ['gh-pages']
   pull_request:
     branches-ignore: ['gh-pages']
-  pull_request_target:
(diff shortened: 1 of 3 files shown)


barbeau (approved):
👍

arriolac:
Looks like the failure is from the test that I just commented out for the `pull_request_target` run. Not sure why it's not picking up the change in this commit. Note though that the other `run-instrumentation-test` runs pass so I believe it should be safe to merge this @wangela 

image

wangela (approved):

PR #203, opened 2022-09-21
Revert "Revert "fix: Prevent a memory leak on AndroidView which may not call ON_DESTROY""

wangela:
Reverts googlemaps/android-maps-compose#202


wangela:
Doesn't preserve credit to contributor. Closing.

PR #206, opened 2022-09-22
fix: Prevent a memory leak on AndroidView which may not call ON_DESTROY

### LINKED ISSUE #138: Memory leak

polivmi1:
This PR fixes a memory leak that can happen in AndroidView when navigating to another screen.
The Lifecycle ON_DESTROY event is never called because in onDispose we remove the mapLifecycleObserver and this method is called before the Lifecycle ON_DESTROY event. It will correct the problem that occurred in https://github.com/googlemaps/android-maps-compose/pull/188

We have to handle any Lifecycle ON_DESTROY in the onDispose method. 

---

Fixes #138  🦕

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -9,7 +9,7 @@ buildscript {
         mavenCentral()
     }
     dependencies {
-        classpath "com.android.tools.build:gradle:7.0.4"
+        classpath 'com.android.tools.build:gradle:7.2.0'
         classpath "com.google.android.libraries.mapsplatform.secrets-gradle-plugin:secrets-gradle-plugin:2.0.1"
         classpath "org.jetbrains.kotlin:kotlin-gradle-plugin:$kotlin_version"
         classpath 'org.jetbrains.dokka:dokka-gradle-plugin:1.5.0'
--- a/gradle/wrapper/gradle-wrapper.properties
+++ b/gradle/wrapper/gradle-wrapper.properties
@@ -1,6 +1,6 @@
 #Fri Dec 03 15:01:34 PST 2021
 distributionBase=GRADLE_USER_HOME
-distributionUrl=https\://services.gradle.org/distributions/gradle-7.0.2-bin.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-7.3.3-bin.zip
 distributionPath=wrapper/dists
 zipStorePath=wrapper/dists
 zipStoreBase=GRADLE_USER_HOME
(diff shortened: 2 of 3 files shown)


wangela (approved):

wangela:
v2.7.2 containing this fix has been released to Maven. Thank you for your contribution @polivmi1!

PR #209, opened 2022-09-27
feat: Add support for StreetView.

### LINKED ISSUE #177: Add Street View Composable
I could not find a JetPack Compose documentation nor code for working with the Google Maps Stretview. I only see Fragments or old school views.  Can you provide a Compose StreetView component? 

Thanks

arriolac:
Fixes #177 

* [x] Support all properties on `StreetViewPanorama`
* [x] Add sample
* [x] Add tests

### DIFF
--- a/README.md
+++ b/README.md
@@ -155,7 +155,7 @@ MarkerInfoWindow(
 
 ### Street View
 
-You can a Street View given a location using the `StreetView` composable.
+You can add a Street View given a location using the `StreetView` composable.
 To use it, provide a `StreetViewPanoramaOptions` object as follows:
 
 ```kotlin
--- a/README.md
+++ b/README.md
@@ -127,7 +127,7 @@ GoogleMap(
 }
 ```
 
-#### Customizing a marker's info window
+### Customizing a marker's info window
 
 You can customize a marker's info window contents by using the 
 `MarkerInfoWindowContent` element, or if you want to customize the entire info 
@@ -153,7 +153,21 @@ MarkerInfoWindow(
 }
 ```
 
-#### Obtaining Access to the raw GoogleMap (Experimental)
+### Street View
+
+You can a Street View given a location using the `StreetView` composable.
+To use it, provide a `StreetViewPanoramaOptions` object as follows:
+
+```kotlin
+val singapore = LatLng(1.35, 103.87)
+StreetView(
+    streetViewPanoramaOptionsFactory = {
+        StreetViewPanoramaOptions().position(singapore)
+    }
+)
+```
+
+### Obtaining Access to the raw GoogleMap (Experimental)
 
 Certain use cases require extending the `GoogleMap` object to decorate / augment
 the map. For example, while marker clustering is not yet supported by Maps Compose
--- a/app/src/main/java/com/google/maps/android/compose/StreetViewActivity.kt
+++ b/app/src/main/java/com/google/maps/android/compose/StreetViewActivity.kt
@@ -1,6 +1,7 @@
 package com.google.maps.android.compose
 
 import android.os.Bundle
+import android.util.Log
 import androidx.activity.ComponentActivity
 import androidx.activity.compose.setContent
 import androidx.compose.foundation.background
@@ -11,14 +12,19 @@ import androidx.compose.foundation.layout.Spacer
 import androidx.compose.foundation.layout.fillMaxSize
 import androidx.compose.foundation.layout.fillMaxWidth
 import androidx.compose.foundation.layout.padding
+import androidx.compose.foundation.lazy.LazyColumn
+import androidx.compose.foundation.lazy.LazyRow
+import androidx.compose.material.Button
 import androidx.compose.material.Switch
 import androidx.compose.material.Text
 import androidx.compose.material.TextField
 import androidx.compose.runtime.Composable
+import androidx.compose.runtime.LaunchedEffect
 import androidx.compose.runtime.getValue
 import androidx.compose.runtime.mutableStateOf
 import androidx.compose.runtime.remember
 import androidx.compose.runtime.setValue
+import androidx.compose.runtime.snapshotFlow
 import androidx.compose.ui.Alignment
 import androidx.compose.ui.Modifier
 import androidx.compose.ui.graphics.Color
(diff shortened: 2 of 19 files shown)


Claveyo1:
Mándame todo de nuevo recuperar mis mapas

lhmzhou (approved):

app/src/main/java/com/google/maps/android/compose/StreetViewActivity.kt
```diff
@@ -0,0 +1,100 @@
...
+            val camera = rememberStreetViewCameraPositionState()
+            LaunchedEffect(camera) {
+                snapshotFlow { camera.panoramaCamera }
+                    .collect {
+                        Log.d(TAG, "Camera at: $it")
+                    }
+                snapshotFlow { camera.location }
+                    .collect {
+                        Log.d(TAG, "Location at: $it")
+                    }
```
DSteve595, line 57: I wouldn't expect this second collection to ever start, since the collection of the first will be suspending. Maybe `launch` these? Or just make two `LaunchedEffect`s and skip the `snapshotFlow`s

maps-compose/src/main/java/com/google/maps/android/compose/streetview/StreetView.kt
```diff
@@ -0,0 +1,145 @@
...
+public fun StreetView(
+    modifier: Modifier = Modifier,
+    cameraPositionState: StreetViewCameraPositionState = rememberStreetViewCameraPositionState(),
+    streetViewPanoramaOptionsFactory: () -> StreetViewPanoramaOptions = {
+        StreetViewPanoramaOptions()
+    },
+    isPanningGesturesEnabled: Boolean = true,
+    isStreetNamesEnabled: Boolean = true,
+    isUserNavigationEnabled: Boolean = true,
+    isZoomGesturesEnabled: Boolean = true,
```
DSteve595, line 43: Purely style-related, but it seems to be convention with Composables in other Google libraries to omit the "is", e.g. `streetNamesEnabled`.

maps-compose/src/main/java/com/google/maps/android/compose/streetview/StreetView.kt
```diff
@@ -0,0 +1,145 @@
...
+    onClick: (StreetViewPanoramaOrientation) -> Unit = {},
+    onLongClick: (StreetViewPanoramaOrientation) -> Unit = {},
+) {
+    val context = LocalContext.current
+    val streetView =
+        remember { StreetViewPanoramaView(context, streetViewPanoramaOptionsFactory()) }
```
DSteve595, line 49: This could leak a Context, since only the `Context` in the first composition will be used. Same problem exists elsewhere in the library.

DSteve595, line 49: How come the panorama options can only be set once? Something preventing them from being updated mid-use?

app/src/main/java/com/google/maps/android/compose/StreetViewActivity.kt
arriolac: Thanks! Going to use `launch` for both of these. I'd prefer to use `snapshotFlow` here since those values can change quite frequently

maps-compose/src/main/java/com/google/maps/android/compose/streetview/StreetView.kt
arriolac: Good call out. I think it makes sense to keep these names as-is for consistency across this library.

maps-compose/src/main/java/com/google/maps/android/compose/streetview/StreetView.kt
arriolac: Unfortunately this is a restriction with `StreetViewPanoramaView`, an options object can only be provided at construction time. The same pattern is observed for `GoogleMapsOptions` https://developers.google.com/android/reference/com/google/android/gms/maps/StreetViewPanoramaView#StreetViewPanoramaView(android.content.Context,%20com.google.android.gms.maps.StreetViewPanoramaOptions)

arriolac: Thanks - yeah this should key against `Context`

DSteve595: Why not use the Context given by AndroidView's `factory`?

arriolac: Thanks. Yeah that's the correct one to use

DSteve595: Ah I see, lame

arriolac: Actually... We can't use that as the `StreetViewPanoramaView` has to be constructed outside of the `AndroidView`'s factory lambda since the `StreetViewPanorama`. It's a bit of a limitation currently with `AndroidView` and breaks guidance on not `remember`ing a View. However, there's no way around this given that it has to be hoisted up.

app/src/main/java/com/google/maps/android/compose/StreetViewActivity.kt
DSteve595: Agreed

wangela (changes requested):
I think docs are generated automatically by bot, but would you add a section to the README about this new feature?

arriolac:
> I think docs are generated automatically by bot, but would you add a section to the README about this new feature?

Done in [e300748](https://github.com/googlemaps/android-maps-compose/pull/209/commits/e30074819ca7b904717a5a04f0e0d353bb6f56ef)

README.md
```diff
@@ -153,7 +153,21 @@ MarkerInfoWindow(
 }
 ```
 
-#### Obtaining Access to the raw GoogleMap (Experimental)
+### Street View
+
+You can a Street View given a location using the `StreetView` composable.
```
wangela, line 158: missing verb. Did you mean "You can add"?

arriolac: Wups. Done

wangela (approved):

PR #213, opened 2022-10-20
fix: `GoogleMap` in preview early return breaking preview layout

### LINKED ISSUE #212: GoogleMap Composable breaks layout in Preview
A previous issue #128 and pull request #179 fix the Compose preview breaking when using a `GoogleMap` composable.
But the layout in the preview is still incorrect as any of the modifiers passed to the `GoogleMap` are discarded with the early return.

For example:

```kotlin
Column(Modifier.fillMaxSize()) {
    GoogleMap(Modifier.weight(1f))
    SomeExampleContent()
}
```

result

A possible fix could be instead of a bare return when in preview, `GoogleMap` could return a `Box` that keeps the modifiers.
```kotlin
// When in preview, early return a Box with the received modifier preserving layout
if (LocalInspectionMode.current) {
    return Box(modifier = modifier)
}
```

I have a commit (7e103f584cf44bda8ceff47706e2578d811d943d) ready to go, but the pull request said to make an issue first.

Thanks

claytongreen:
Fixes #212

### DIFF
--- a/maps-compose/src/main/java/com/google/maps/android/compose/GoogleMap.kt
+++ b/maps-compose/src/main/java/com/google/maps/android/compose/GoogleMap.kt
@@ -90,7 +90,8 @@ public fun GoogleMap(
 ) {
     // When in preview, early return a Box with the received modifier preserving layout
     if (LocalInspectionMode.current) {
-        return Box(modifier = modifier)
+        Box(modifier = modifier)
+        return
     }
 
     val context = LocalContext.current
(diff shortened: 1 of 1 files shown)


wangela (approved):

arriolac (approved):
Change LGTM. The test failures is due to the Maps API key secret not being copied over from forked PRs

wangela:
Thank you for your contribution @claytongreen!

PR #224, opened 2022-11-17
chore: Upgrade Gradle to 7.5.1, AGP to 7.3.1

DSteve595:
Maintenance. Hopefully will bring improvements to Managed Devices so we can use those for connected tests.


DSteve595:
Aww.

arriolac (approved):

PR #235, opened 2022-12-08
perf: Make MarkerState stable

### LINKED ISSUE #232: MarkerState behaves in a non-@Stable manner at runtime
Maps Compose 2.7.2

Using `MarkerState` as a Composable function parameter appears to render the function effectively non-skippable at runtime. I suspect that this is due to unstable `LatLng` `MarkerState.position`.

I'd think that explicitly marking `MarkerState` `@Stable` would easily and validly avoid this issue.

The function below recomposes whenever its parent recomposes, even when keeping MarkerState unchanged:

```
@Composable
fun Foo(markerState: MarkerState) {
    println("${markerState.position} - ${markerState.dragState}")
}
```

DSteve595:
`MarkerState`'s `internal var: Marker?` property was causing the class to be marked unstable.
While I think it can safely be marked `@Stable` as-is, I'd rather have the compiler infer that. I feel better backing it with an actual state value, that way if we later do an actual non-imperative state read it won't break stuff.

Fixes #232 🦕

### DIFF
--- a/maps-compose/src/main/java/com/google/maps/android/compose/Marker.kt
+++ b/maps-compose/src/main/java/com/google/maps/android/compose/Marker.kt
@@ -18,6 +18,7 @@ import androidx.compose.runtime.Composable
 import androidx.compose.runtime.ComposeNode
 import androidx.compose.runtime.CompositionContext
 import androidx.compose.runtime.Immutable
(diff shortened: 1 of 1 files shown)


wangela (approved):

DSteve595:
Superseded by #254

PR #236, opened 2022-12-08
perf: Make CameraPositionState stable

DSteve595:
`CameraPositionState` has a few unstable properties. Rather than verify all usages of these are safe for stability, I'm opting to back them with mutable states, so the compiler can mark the class stable for us.


DSteve595:
Superseded by #254

PR #238, opened 2022-12-12
feat: Expose function for clearing tile overlay tiles

### LINKED ISSUE #197: Force TileOverlay to redraw tiles
For a standard GoogleMap TileOverlay you can call` tileOverlay.clearTileCache()`.

Not sure how to invalidate and force compose TileOverylay to redraw tiles.

Using:
```
GoogleMap(...) {
  ...
  // Add tile overlay with tile provider
  TileOverlay(tileProvider = tileProvider)
  ...
}
```

DSteve595:
I think it would be more Compose-like to observe created Tiles as a state, rather than imperatively invalidate them. However, the underlying Maps API doesn't seem to play nice with that, so this is better than nothing.

Fixes #197 🦕


wangela (approved):
Thanks for improving the usage of TileOverlay

PR #243, opened 2022-12-20
fix: Migrate publication to AGP's publishing API

DSteve595:
This removes most of the custom setup of the Maven publication in favor of [AGP's components](https://developer.android.com/studio/publish-library).

Some notable changes:
- Before, the generated POM did not contain any dependencies. The old code suggests it might, but since it only adds dependencies from the `api` configuration, of which `:maps-compose` has none, the actual `<dependencies/>` block in the POM was empty.
With this change, the POM has expected dependencies, e.g. `play-services-maps`.
- Gradle Module metadata is now present in the publication. To my knowledge, this shouldn't result in any noticeable changes for users, as we aren't shipping test fixtures.

Generated Javadoc jar still contains Dokka stuff.

### DIFF
--- a/README.md
+++ b/README.md
@@ -263,9 +263,6 @@ dependencies {
     // Make sure to also include the latest version of the Maps SDK for Android 
     implementation 'com.google.android.gms:play-services-maps:18.0.2'
     
-    // Also include Compose version `1.2.0-alpha03` or higher - for example:
-    implementation 'androidx.compose.foundation:foundation:2.8.0-alpha03'
-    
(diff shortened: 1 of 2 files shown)


DSteve595:
Changing `chore` to `fix` since this fixes #246.

wangela (changes requested):
Thank you for this fix! Please update the README to tidy up the Installation instructions and remove the unnecessary dependency.

wangela (approved):

PR #244, opened 2022-12-21
chore: Fail tests when API key is missing

DSteve595:
Currently when no API key is provided, the instrumentation tests fail with a cryptic "Map loaded" exception. This adds a more explicit error when no API key is provided.

### DIFF
--- a/app/src/androidTest/java/com/google/maps/android/compose/GoogleMapViewTests.kt
+++ b/app/src/androidTest/java/com/google/maps/android/compose/GoogleMapViewTests.kt
@@ -40,10 +40,7 @@ class GoogleMapViewTests {
     private lateinit var cameraPositionState: CameraPositionState
 
     private fun initMap(content: @Composable () -> Unit = {}) {
-        check(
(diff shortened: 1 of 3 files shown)


app/src/androidTest/java/com/google/maps/android/compose/GoogleMapViewTests.kt
```diff
@@ -40,6 +40,7 @@ class GoogleMapViewTests {
     private lateinit var cameraPositionState: CameraPositionState
 
     private fun initMap(content: @Composable () -> Unit = {}) {
+        check(BuildConfig.MAPS_API_KEY != "YOUR_API_KEY") { "Maps API key not specified" }
```
wangela, line 43: Maybe we should also check that MAPS_API_KEY is not blank, in case for some reason the default placeholder isn't read.

wangela (approved):
Please add a description to the PR, thanks!

app/src/androidTest/java/com/google/maps/android/compose/GoogleMapViewTests.kt
DSteve595: Added!

PR #254, opened 2023-01-18
perf: Make CameraPositionState and MarkerState stable

### LINKED ISSUE #232: MarkerState behaves in a non-@Stable manner at runtime

DSteve595:
`CameraPositionState` has a few unstable properties. Rather than verify all usages of these are safe for stability, I'm opting to back them with mutable states, so the compiler can mark the class stable for us.
`MarkerState`'s `internal var: Marker?` property was causing the class to be marked unstable.
While I think it can safely be marked `@Stable` as-is, I'd rather have the compiler infer that. I feel better backing it with an actual state value, that way if we later do an actual non-imperative state read it won't break stuff.

Fixes #232 🦕

### DIFF
--- a/maps-compose/src/main/java/com/google/maps/android/compose/CameraPositionState.kt
+++ b/maps-compose/src/main/java/com/google/maps/android/compose/CameraPositionState.kt
@@ -110,16 +110,16 @@ public class CameraPositionState(
 
     // Used to perform side effects thread-safely.
     // Guards all mutable properties that are not `by mutableStateOf`.
-    private val lock = Any()
+    private val lock = Unit
 
     // The map currently associated with this CameraPositionState.
     // Guarded by `lock`.
-    private var map: GoogleMap? = null
+    private var map: GoogleMap? by mutableStateOf(null)
 
     // An action to run when the map becomes available or unavailable.
     // represents a mutually exclusive mutation to perform while holding `lock`.
     // Guarded by `lock`.
-    private var onMapChanged: OnMapChangedCallback? = null
+    private var onMapChanged: OnMapChangedCallback? by mutableStateOf(null)
 
     /**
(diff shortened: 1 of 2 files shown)


wangela (approved):

bubenheimer:
Nervous about seeing additional state added to MarkerState: an app may have quite a lot of Marker/MarkerState objects. Were time/space performance impacts considered?

DSteve595:
@bubenheimer No, those weren't considered. A benchmark would be useful. I made the judgment call that stability of a State class is generally expected, and so ours should fit the pattern.
Have you encountered any performance issues from this?

bubenheimer:
The Marker-related change looks like a red flag to me. `Marker` is a mutable type inside what's now a MutableState. And on top of that there is `MarkerState.position`, which replicates parts of the MutableState-wrapped mutable object inside another MutableState.

As you may be aware, I believe that `MarkerState` can and should be eliminated entirely, at least from the public API surface: #149

DSteve595:
I'm gonna take some time to digest your thoughts in #149. Thanks very much for the detailed explanations.

The MarkerState changes were made with defensive correctness in mind (I'm not positive about enumerating all cases that can modify those properties, now and in the future). I'm inclined to try benchmarking some scenarios.

PR #258, opened 2023-01-24
feat: Add Clustering composable, maps-compose-utils library

### LINKED ISSUE #44: RFC: Add support for clustering

DSteve595:
This adds a Compose-friendly API for marker clustering. I'd appreciate heavy scrutiny on it, since it's got some hacks and ugly bits!

`MapClusteringActivity` has been updated to show an example of how it can be used. Notably, the bug with Marker (and other) click handlers no longer working, as described in https://github.com/googlemaps/android-maps-compose/pull/140#discussion_r894716482, is fixed! It's unfortunately done by working around some unwanted behavior of maps-utils, where it clobbers the click listeners that `MapApplier` previously set. To facilitate that, I made some changes to how `MapApplier` dispatches input, namely a new `InputHandler` composable that can be added to arbitrarily handle all unhandled input. This solves the problem for clustering since ClusterManager wants to receive input events directly, and the current implementation throws away the marker click events since we haven't added corresponding `MarkerNode`s for them.

To match the structure of maps-utils (which contains the non-Compose clustering utilities), this also adds a new artifact, maps-compose-utils, in which the Clustering composable lives.

To do before merging:
- [x] Update readme to mention `Clustering` instead of `MapEffect`
- [ ] Update kdoc in `MapEffect`

Fixes #44


DSteve595:
Open question: Do we still need `MapEffect`? Clustering is the only use case I'm aware of for it (though there certainly could be more), and it introduces lots of footguns

StephenVinouze:
Except if I've missed it, I don't see a way to customize the cluster icons. That's critical for branding and one of the pain points when using maps-compose with the former util library since you need to shape your marker and cluster icons with the View system and turn them into bitmaps (preventing us from using our Composables).

DSteve595:
Correct, that isn't accounted for in this PR. Will explore options.

ColtonIdle:
Adding "true" clustering support with compose-maps would be great. I don't mind the MapEffect to be honest since it works, but something more "native" is def fine with me. But to @StephenVinouze point of customizing the marker icons and cluster icons would be like a 100% need.

Good work getting it this far @DSteve595

maps-compose-utils/src/main/java/com/google/maps/android/compose/clustering/Clustering.kt
```diff
@@ -0,0 +1,113 @@
...
+ * @param onClusterItemInfoWindowClick a lambda invoked when the user clicks the info window of a
+ * non-clustered item
+ * @param onClusterItemInfoWindowLongClick a lambda invoked when the user long-clicks the info
+ * window of a non-clustered item
+ */
+@MapsComposeExperimentalApi
+@Composable
+public fun <T : ClusterItem> Clustering(
+    items: Collection<T>,
+    cameraPositionState: CameraPositionState,
```
arriolac, line 40: Suggestion/idea: how about making this a part of the `MapApplier` so that you can access it implicitly?

maps-compose/src/main/java/com/google/maps/android/compose/MapApplier.kt
```diff
@@ -68,94 +68,138 @@ internal class MapApplier(
...
-    groundOverlay: GroundOverlay
-): GroundOverlayNode? =
-    firstOrNull { it is GroundOverlayNode && it.groundOverlay == groundOverlay } as? GroundOverlayNode
+/**
+ * General pattern for handling input:
+ * Find the node that belongs to the clicked item.
+ * If there is none, default to the first InputHandlerNode.
+ * If there is none, don't handle.
+ */
+private inline fun <reified NodeT : MapNode, I, O> Iterable<MapNode>.findInputCallback(
```
arriolac, line 189: 👌

arriolac (commented):
Have you considered implementing `Clustering` as a `ComposeNode<ClusteringNode, MapApplier>`? The advantage of implementing it this way is that you can update underlying state in the same frame that recomposition happens. With `LaunchedEffect`, properties won't be updated until the next frame which means you might lose the click events

maps-compose-utils/src/main/java/com/google/maps/android/compose/clustering/Clustering.kt
DSteve595: Was wondering if that'd be prudent. I think I'll go for it.

maps-compose/src/main/java/com/google/maps/android/compose/MapApplier.kt
DSteve595: Honestly worried this is "too clever", but it certainly solves the problem

arriolac: It's definitely much cleaner!

DSteve595:
> Have you considered implementing `Clustering` as a `ComposeNode<ClusteringNode, MapApplier>`? The advantage of implementing it this way is that you can update underlying state in the same frame that recomposition happens. With `LaunchedEffect`, properties won't be updated until the next frame which means you might lose the click events

I wanted to directly reference `MapApplier` for a few things, but since this is technically in another library (maps-compose-utils), I don't have access to it (it's internal). Forced to eat our own MapEffect dogfood.

Can you give an example of how a click event might be lost? I'm assuming you're referring to the LaunchedEffect that sets the click listeners. As I understand it, the user would have to click a marker within a single choreographer frame after an `onClick` lambda is changed. I don't think that's even possible, since clicks take multiple frames of input to process.

arriolac:
> @arriolac
> 
> > Have you considered implementing `Clustering` as a `ComposeNode<ClusteringNode, MapApplier>`? The advantage of implementing it this way is that you can update underlying state in the same frame that recomposition happens. With `LaunchedEffect`, properties won't be updated until the next frame which means you might lose the click events
> 
> I wanted to directly reference `MapApplier` for a few things, but since this is technically in another library (maps-compose-utils), I don't have access to it (it's internal). Forced to eat our own MapEffect dogfood.
> 
> Can you give an example of how a click event might be lost? I'm assuming you're referring to the LaunchedEffect that sets the click listeners. As I understand it, the user would have to click a marker within a single choreographer frame after an `onClick` lambda is changed. I don't think that's even possible, since clicks take multiple frames of input to process.

Ahh, that's right, this is a different module altogether. 

I was thinking an event can be lost in this scenario:
* state triggers recomposition and you update the click handler 
* click event occurs
* new handler hasn't been applied yet since the block inside LaunchedEffect hasn't been executed yet

It's an unlikely edge case I thought worth calling out though it would definitely have to be validated if that's even possible. Using `ComposeNode` would rule this out entirely but I'm hesitant about changing the scope of `MapApplier` for this use case.

maps-compose-utils/src/main/java/com/google/maps/android/compose/clustering/Clustering.kt
DSteve595: Leaning towards a non-providable local, to avoid giving MapApplier even more responsibilities.

DSteve595:
I updated the PR with a new arg to `Clustering`: `clusterContent: (@Composable (Cluster) -> Unit))?`. This lets you render custom cluster icons using composables. Marker icons will follow shortly, but in the meantime I'd appreciate a look.
This also contains two other PRs, #261 and #262, which enable some of the functionality needed.

DSteve595:
@arriolac I've been trying to come up with a test case that repros missing click handler events, but I've been coming up dry. I might be missing a case you're imagining. If you find the time, would you be able to come up with one?
I'm not considering it a blocker currently because the API itself should be fine, we can always patch the implementation.

maps-compose-utils/src/main/java/com/google/maps/android/compose/clustering/Clustering.kt
DSteve595: Actually, a public Composable property seems better, since then I can annotate it with `@GoogleMapComposable`.

arriolac:
> @arriolac I've been trying to come up with a test case that repros missing click handler events, but I've been coming up dry. I might be missing a case you're envisioning. If you find the time, would you be able to come up with one? I'm not considering it a blocker currently because the API itself should be fine, we can always patch the implementation.

Agreed that it shouldn't be blocking as the underlying implementation can change. An alternative to using `LaunchedEffect` in https://github.com/googlemaps/android-maps-compose/pull/258/files#diff-73d069b3c67a66f9bbaf70ba78d487309d39454a28940f02898176dabd6e6414R73 that doesn't have the downsides I mentioned is to use `SideEffect` as the lambda inside will be invoked in the same frame as the recomposition.

DSteve595:
Note: Exploring possibilities for removing the no-animation restriction, since that also would break async image loaders

DSteve595:
With some invalidation hacks, marker/cluster icons are now rendered continuously! This importantly fixes async image loaders (like Glide), and also allows for animations.

wangela (approved):

wangela:
Noting here some decisions we  have made in this PR:

- Add a `maps-compose-utils` library even though it only contains the one utility for now, Marker Clustering.
- Manage the code for the maps-compose-utils library in this repo but distribute it as a separate dependency through Maven, similar to `maps-compose-widgets`, to be close to the code for the related composables.
- Keep documentation about the Experimental `MapEffect`, in case direct control of the map is needed for non-clustering functionality. Remove documentation only when `MapEffect` is removed (#267).

ColtonIdle:
Great job @DSteve595 Can't wait to use this!

PR #284, opened 2023-03-08
docs: update README

wangela:


Alexs784:
Regarding the clustering part: something I would mention and that was not clear (at least to me), is that the `com.google.maps.android:android-maps-utils:x.y.z` dependency is needed (because of the `ClusterItem` interface)

wangela:
> Regarding the clustering part: something I would mention and that was not clear (at least to me), is that the `com.google.maps.android:android-maps-utils:x.y.z` dependency is needed (because of the `ClusterItem` interface)

Thanks for pointing this out! I've filed #288 and there's a PR in progress to clarify this.

PR #287, opened 2023-03-10
fix: Expose maps-utils as API dependency from maps-compose-utils

DSteve595:
This changes `maps-utils` to be an API dependency, that way users don't have to add it manually.


wangela (approved):

wangela:
Is this considered a breaking change? i.e. if people have also added `android-maps-utils` or `maps-utils-ktx` as a dependency will they see problems when this seeks to load `maps-utils-ktx` as an API dependency, and need to remove their existing dependencies?

DSteve595:
Not a breaking change. If someone already has the dependency declared explicitly, it will continue to work.

PR #296, opened 2023-03-27
Fix Compose 1.4 layout issue 

### LINKED ISSUE #295: Layout using GoogleMap breaks since Compose 1.4.0
I just noticed when updating from Compose BOM `2023.01.00` (Compose `1.3.3`) to `2023.03.00` (Compose `1.4.0`), a layout with a `GoogleMap` composable breaks.

I have this (simplified) layout:

```kotlin
Column {
  GoogleMap(
    modifier = Modifier.fillMaxWidth().weight(1f)
  ) {
    …
  }

  Box {
    // More contents
  }
}
```

so `GoogleMap` should only take the remaining space. This worked fine before but now the `GoogleMap` composable fills the whole screen and the contents of `Box` is not visible anymore.

I'm not sure if this is some incompatibility with Compose 1.4.0?

Using `maps-compose` version `2.11.3`.

svenjacobs:
This pull request fixes #295 and also updates Compose dependencies.

If an update of Compose dependencies in the same PR is not welcomed, the second commit can just be dropped.


svenjacobs:
Closing because this is a [Compose issue](https://issuetracker.google.com/issues/274797771).

PR #308, opened 2023-04-17
Add StyleSpans support for Polylines

### LINKED ISSUE #307: Support of StyleSpan in Polylines
It would be nice to have support for StyleSpan in polylines, as in the non-compose version. That would allow us to leverage the new renderer capabilities to draw multicoloured polylines including gradient segments.

In my case, every polyline segment needs to have a  different colour. Therefore, to achieve desired behaviour I need to add lots of 2-point polylines. That is not very efficient. Also, when the polylines are not straight, it is not smooth and a user can see visible sharp disconnected edges.

Add support for setting StyleSpan in PolylineOptions for each segment.

lenoch0d:
Fixes #307 🦕

- updated google play maps services which support the new feature
- added new `Polyline` composable with the `spans` parameter, allowing to specify spans for each segment of `points`
- I did not extend the existing `Polyline` composable, because I believe it would be confusing with the `color` parameter - it doesn't make sense to specify it when we specify Spans for every segment. To not make it confusing, I think we'd need to make it nullable without the default color. However, that would be a braking change.

### DIFF
--- a/maps-compose/src/main/java/com/google/maps/android/compose/Polyline.kt
+++ b/maps-compose/src/main/java/com/google/maps/android/compose/Polyline.kt
@@ -25,6 +25,7 @@ import com.google.android.gms.maps.model.JointType
 import com.google.android.gms.maps.model.LatLng
 import com.google.android.gms.maps.model.PatternItem
 import com.google.android.gms.maps.model.Polyline
+import com.google.android.gms.maps.model.StyleSpan
 import com.google.maps.android.ktx.addPolyline
 
 internal class PolylineNode(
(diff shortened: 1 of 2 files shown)


wangela (approved):

wangela (changes requested):
Thank you for this contribution! The implementation looks good and this will be useful for polyline customization.

- Please add tests.
- Please add example usage to the demo app and README.
- You may want to rebase or merge from main.

maps-compose/src/main/java/com/google/maps/android/compose/Polyline.kt
```diff
@@ -106,4 +107,76 @@ public fun Polyline(
...
+            set(geodesic) { this.polyline.isGeodesic = it }
+            set(jointType) { this.polyline.jointType = it }
+            set(pattern) { this.polyline.pattern = it }
+            set(startCap) { this.polyline.startCap = it }
+            set(tag) { this.polyline.tag = it }
+            set(visible) { this.polyline.isVisible = it }
+            set(width) { this.polyline.width = it }
+            set(zIndex) { this.polyline.zIndex = it }
+        }
+    )
```
bubenheimer, line 181: Can you create a private `PolylineImpl()` function to combine the nearly identical implementations of the two `Polyline()` overloads?

bubenheimer (changes requested):

el-qq:
Hi! 
I wanted to check in and see if you're planning to continue working on it? We're looking forward to potentially incorporating these changes into our project.

el-qq:
@kikoso  Could you please close the PR? (outdated)

Issue #38, opened 2022-02-16
Suggestion: don't set default map click listeners other than OnMapClickListener


fbarthelery:
`MapView` use a layer approach to handle click on map element. If a click is not handled by a high layer, a lower layer will be able to handle it. For exemple, when a click on a POI occurs, the POI layer will invoke `OnPoiClickListener` if it is set. If it isn't set the `OnMapClickListener` will be called.
(Note: I'm not sure of the exact implementation details, I just described what I think it is)

`GoogleMap()` composable always set   `OnPoiClickListener`,  `OnMyLocationClickListener`,  so the fallback to `OnMapClickListener` and the associate `onMapClick` callback is never called and you are forced to implement `onPOIClick` to handle the click even if you don't care about the POI.  MapView also has a default behavior for the onMapClick if you don't set `OnMapClickListener` that you may want to preserve

jpoehnelt:
@fbarthelery Thank you for opening this issue. 🙏
Please check out these other resources that might be applicable:

- [Support Console](https://cloud.google.com/support/) - If you have a support contract.
- [Discord](https://discord.gg/9UpV862xZ9) - chat with other developers
- [StackOverflow](http://stackoverflow.com/questions/tagged/google-maps) - use the `google-maps` tag
- [Google Maps Platform issue trackers](https://developers.google.com/maps/support/#issue_tracker)


_This is an automated message, feel free to ignore._

arriolac:
Thanks for reporting this issue. Maps Compose should mirror the same behavior as the underlying SDK so this should definitely be integrated.

romainpiel:
I think even `OnMapClickListener` should not be set by default. The current implementation prevents clicks on a parent composable.

For example:

```kotlin
Box(Modifier.clickable { foo() }) {
  GoogleMap()
}
```

`foo()` will never be called on a click

wangela:
Thanks @fbarthelery for making this suggestion and @romainpiel for offering a fix! We're validating some of the behaviors reported here.

Tetr4:
Hi, this is also a usability issue, because it prevents users from clicking on markers they are standing directly in front of. `onMyLocationClick` is not nullable, so it always sets a `OnMyLocationClickListener` on the underlying `MapView`, which causes the "my location" layer to intercepts clicks.

PR #320, opened 2023-05-23
fix: Make all listeners null by default

romainpiel:
Prior to this change, all click listeners and onMapLoaded were set to an empty lambda by default. This would cause issues where for example the map is clickable by default without doing anything. When talkback is on, it reads "double click to activate".

This change fixes that by making all listeners null by default.

---

Fixes https://github.com/googlemaps/android-maps-compose/issues/38 🦕


wangela (approved):

wangela:
Thank you for this fix, @romainpiel! Marking this a chore so it gets bundled with a couple of other changes in a planned release.

PR #327, opened 2023-06-12
fix: Set initial map padding (#326)

### LINKED ISSUE #326: Content padding is not used for initial rendering
When you set the `contentPadding` parameter, it is not considered when the map is initially rendered. This results in `cameraPositionState.position` being off-center. Subsequent updates will correctly center taking the content padding into account.

I created a simple [reproducing example](https://github.com/kkris/android-maps-compose/tree/missing-initial-content-padding-demo) here: https://github.com/kkris/android-maps-compose/commit/865bd3c9d167fffb862ccacdb018cadde57512a3

Here is a video of this bug: https://github.com/googlemaps/android-maps-compose/assets/232214/ed5b164f-d926-4ca7-afdb-ffb14f3fe4e1

kkris:
Fixes #326 🦕

With this fix applied, the example from #326 now behaves correctly: https://github.com/googlemaps/android-maps-compose/assets/232214/7a07bcbb-c9fc-4924-8131-762926f36e29


kkris:
@kikoso Thanks for reviewing! Any idea how to fix the `Run instrumentation tests` check?

```
java.lang.IllegalStateException: Maps API key not specified
	at com.google.maps.android.compose.MapInColumnTests.initMap(MapInColumnTests.kt:43)
```

maps-compose/src/main/java/com/google/maps/android/compose/MapUpdater.kt
```diff
@@ -152,15 +155,7 @@ internal inline fun MapUpdater(
         set(mapProperties.maxZoomPreference) { map.setMaxZoomPreference(it) }
         set(mapProperties.minZoomPreference) { map.setMinZoomPreference(it) }
         set(contentPadding) {
```
bubenheimer, line 157: This needs to change from `set()` to `update()`. Probably should move up to the other `update()` calls purely for consistency.

```suggestion
        update(contentPadding) {
```

bubenheimer (changes requested):

maps-compose/src/main/java/com/google/maps/android/compose/MapUpdater.kt
```diff
@@ -34,9 +34,11 @@ internal class MapPropertiesNode(
     var clickListeners: MapClickListeners,
     var density: Density,
     var layoutDirection: LayoutDirection,
+    contentPadding: PaddingValues
 ) : MapNode {
 
     init {
+        applyContentPadding(map, contentPadding)
         cameraPositionState.setMap(map)
```
bubenheimer, line 42: Could you add a brief comment here to inform a reader why padding must precede camera position?

```suggestion
        applyContentPadding(map, contentPadding)
        // set camera position after padding for correct centering
        cameraPositionState.setMap(map)
```

kkris:
I rebased the branch on current main and applied the review comments.

PR #347, opened 2023-07-23
feat: customizing the location button

### LINKED ISSUE #340: Location button is not customizable
Current location button is not customizable and is not supporting overriding of the it's onClick method. There can be multiple use cases where the location should redirect the map to a specific location, not the current device location.

kikoso:
The follogin PR adds customization options to the location button. We can add our own Composable as a custom Location button, or modify only the click behavior

Fixes #340 🦕

### DIFF
--- a/app/src/main/java/com/google/maps/android/compose/CustomLocationButtonActivity.kt
+++ b/app/src/main/java/com/google/maps/android/compose/CustomLocationButtonActivity.kt
@@ -1,0 +1,109 @@
+// Copyright 2023 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package com.google.maps.android.compose
+
+import android.os.Bundle
+import android.widget.Toast
+import androidx.activity.ComponentActivity
+import androidx.activity.compose.setContent
+import androidx.compose.animation.AnimatedVisibility
+import androidx.compose.animation.EnterTransition
+import androidx.compose.animation.fadeOut
+import androidx.compose.foundation.background
+import androidx.compose.foundation.layout.Box
+import androidx.compose.foundation.layout.fillMaxSize
+import androidx.compose.foundation.layout.padding
+import androidx.compose.foundation.layout.wrapContentSize
+import androidx.compose.material.Button
+import androidx.compose.material.ButtonDefaults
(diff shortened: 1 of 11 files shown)


vladcioricagoogle:
Hi @kikoso , yes I would say that we should have a dedicated param for location button and one also for zooming controls. Actually I've added it through the content param but for example is has to stay somewhere over the zooming controls. To match their exact padding I set it's bottom padding to 100.dp and it's right padding to 7.dp, which feels pretty wrong to use that magic numbers just because it looks good. I feel that it will be more easier to use it like this and cutomize it directly from a dedicated param. 
What are your thoughts for this?

vladcioricagoogle:
This LGTM. I'll wait for merging this CL and then I'll try to do the same for zooming controls. These one's don't even need functionality overriding.

app/src/main/java/com/google/maps/android/compose/CustomLocationButtonActivity.kt
wangela, line 1: Let's name this CustomControlsActivity so that the future custom Zoom controls (and any other future map controls) can be added to it rather than creating many separate activities.

app/src/main/res/values/strings.xml
```diff
@@ -23,4 +23,5 @@
   <string name="location_tracking_activity">Location Tracking</string>
   <string name="scale_bar_activity">Scale Bar</string>
   <string name="street_view">Street View</string>
+  <string name="custom_location_button">Custom Location Button</string>
```
wangela, line 26: `custom_my_location_button` and "Custom My Location Button" to match the full name of the control ("My Location")

app/src/main/java/com/google/maps/android/compose/LocationTrackingActivity.kt
```diff
@@ -100,6 +115,7 @@ class LocationTrackingActivity : AppCompatActivity() {
                     onMapLoaded = {
                         isMapLoaded = true
                     },
+                    onMyLocationButtonClick = {  Log.d(TAG,"Overriding the onMyLocationButtonClick with this Log"); true },
```
wangela, line 118: This line could benefit from a comment indicating that it's demonstrating use of a Custom My Location Button. Otherwise people might end up leaving it empty and unwittingly disable the default My Location Button.

wangela (changes requested):
Thank you for this implementation!

vladcioricagoogle:
Is this ready to be merged? @kikoso @wangela

emmano:
@kikoso Thanks for taking the time to do this. Any updates on when it will be merged?

wangela (approved):

wangela (approved):

wangela:
@kikoso can you look into why instrumentation tests are failing? Error:

```
com.google.maps.android.compose.MapInColumnTests > testPanMapUp_MapCameraChangesColumnDoesNotScroll[test(AVD) - 10] FAILED 
	java.lang.AssertionError: Failed to inject touch input.
	Reason: Expected exactly '1' node but found '2' nodes that satisfy: (TestTag = 'Map')
Tests on test(AVD) - 10 failed: There was 1 failure(s).
```

PR #353, opened 2023-07-27
feat: added Factory for Map

### LINKED ISSUE #342: Add a factory map to be used in tests
With current implementation the google map with jetpack compose can't be UI tested. It should provide a map for tests in it's constructor.

kikoso:
The following PR creates a Factory for GoogleMap, and it facilitates its usage on tests.

---

Fixes #342 🦕


vladcioricagoogle:
I would say so, thanks @kikoso! Only one question, what are you using the "singapore" related variables for? It seems like you just added there a marker state but didn't do anything with it.

wangela (approved):

Issue #380, opened 2023-08-17
Huge drop in performances when setting custom clusterContent or clusterItemContent with many items


AdrienDruesne:
Hello,

While using the Clustering in the `com.google.maps.android:maps-compose-utils:2.13.0` library, if I customize the clusterContent or clusterItemContent, I have a huge drop in performances with 100+ items.
Even for a simple composable as the one below, the app stutter when I move on the map, but I don't have any problem if I leave clusterContent and clusterItemContent to null.

         clusterContent = { cluster ->
                 Surface(
                     Modifier.size(40.dp),
                     shape = CircleShape,
                     color = Color.Blue,
                     contentColor = Color.White,
                     border = BorderStroke(1.dp, Color.White)
                 ) {
                     Box(contentAlignment = Alignment.Center) {
                         Text(
                             "%,d".format(cluster.size),
                             fontSize = 16.sp,
                             fontWeight = FontWeight.Black,
                             textAlign = TextAlign.Center
                         )
                     }
                 }
             },

Thanks!

wangela:
If you would like to upvote the priority of this issue, please comment below or react on the original post above with :+1: so we can see what is popular when we triage.
--------

@AdrienDruesne Thank you for opening this issue. 🙏
Please check out these other resources that might help you get to a resolution in the meantime:

- [Check the issue tracker](https://developers.google.com/maps/support/#issue_tracker) - bugs and feature requests for Google Maps Platform APIs and SDKs
- [Open a support case](https://developers.google.com/maps/support#contact-maps-support) - Get 1:1 support in Cloud Console.
- [Discord](https://discord.gg/fHTDmZe9En) - chat with other developers
- [StackOverflow](http://stackoverflow.com/questions/tagged/google-maps) - use the `google-maps` tag


_This is an automated message, feel free to ignore._

cwsiteplan:
Same here, it's making the app freeze and being unusable at the moment.
But even when using the default cluster+item config it's not working when adding 1000+ (clustered) markers

cwsiteplan:
to replicate - just increase the cluster count of the ClusterSample to e.g. 4000
lot's of GC happening according to logs and screen is freezing.

cwsiteplan:
would it be possible to expand clusters only for the visible viewport? - seems like it's currently expanding on zoom level - adding lot's of individual markers on areas that are potentially not viewed at all

wangela:
[...]

PR #421, opened 2023-10-04
feat: Implemented so that ClusterManager#setAlgorithm can be use.

### LINKED ISSUE #420: Allows use of ClusterManager#setAlgorithm
In Compose, ClusterManager is hidden internally, so `ClusterManager#setAlgorithm` cannot be used.

Make this available.

I think it can be implemented in the following below ways. 

- Pass the algorithm to Clustering Composable.
    - In the future, when all the functions available in Clustering are implemented, the number of arguments may increase.
-  Add the `clusterManagerFactory: () -> ClusterManager<T>` argument to Clustering Composable and create it.
    - You can do everything you can do with View.
    - Bad. `ClusterManager#renderer` has been changed internally, causing confusion for users.
- Call the Clustering Composable with the argument  `onCreateClusterManager: (ClusterManager<T>) -> Unit` 
    - You can do everything you can do with View.
    - We believe this will be less confusing than the above because the user rewrites the behavior.
    - However, even in this case, if clusterRenderer is not specified, the behavior is internally overridden.
        - https://github.com/googlemaps/android-maps-compose/blob/8c6c3b0ae14b5325b552bbcafa093372c2448c60/maps-compose-utils/src/main/java/com/google/maps/android/compose/clustering/Clustering.kt#L116-L118
- [recommend] create rememberClusterRenderer with ClusterManager as an argument.
    - Currently, clusterContent and clusterItemContent are not used when clusterRenderer is specified. I think this is complicated.
   - Prepare a low-level API using ClusterManager that can do the same thing as the View, and a high-level API.

 **Additional context**
Our app has a large number of pins and we need to use `NonHierarchicalViewBasedAlgorithm` of  `ScreenBasedAlgorithm`.

matsudamper:
Existing problems
- ClusterManager#setAlgorithm is not available.
- clusterContent and clusterItemContent are not used when clusterRenderer is specified.

As described in the Issue, we thought it would be less confusing for users to separate the low-level API and the high-level API.

Clustering Composable Function that takes a clusterManager as an argument, memberClusterRenderer and memberClusterManager are implemented and make public.
The existing Clustering Composable Function calls the low-level Clustering Composable Function.

-----

Fixes #420 🦕

### DIFF
--- a/app/src/main/java/com/google/maps/android/compose/MarkerClusteringActivity.kt
+++ b/app/src/main/java/com/google/maps/android/compose/MarkerClusteringActivity.kt
@@ -5,16 +5,29 @@ import android.util.Log
 import androidx.activity.ComponentActivity
 import androidx.activity.compose.setContent
 import androidx.compose.foundation.BorderStroke
+import androidx.compose.foundation.ScrollState
+import androidx.compose.foundation.horizontalScroll
+import androidx.compose.foundation.layout.Arrangement
 import androidx.compose.foundation.layout.Box
+import androidx.compose.foundation.layout.Row
 import androidx.compose.foundation.layout.fillMaxSize
+import androidx.compose.foundation.layout.fillMaxWidth
+import androidx.compose.foundation.layout.padding
 import androidx.compose.foundation.layout.size
 import androidx.compose.foundation.shape.CircleShape
+import androidx.compose.material.Button
+import androidx.compose.material.ButtonDefaults
+import androidx.compose.material.MaterialTheme
 import androidx.compose.material.Surface
 import androidx.compose.material.Text
 import androidx.compose.runtime.Composable
 import androidx.compose.runtime.LaunchedEffect
+import androidx.compose.runtime.SideEffect
+import androidx.compose.runtime.getValue
 import androidx.compose.runtime.mutableStateListOf
+import androidx.compose.runtime.mutableStateOf
 import androidx.compose.runtime.remember
+import androidx.compose.runtime.setValue
 import androidx.compose.ui.Alignment
 import androidx.compose.ui.Modifier
 import androidx.compose.ui.graphics.Color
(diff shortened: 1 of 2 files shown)


matsudamper:
There are 10 icons set, but 11 are displayed at main branch. I'm looking into whether the issues are related.

matsudamper:
Even if you branch Clustering with if, the marker remains displayed. This issue may be related.

matsudamper:
fixed.

<table>
<tr>
<td></td>
<td></td>
<td></td>
</tr>
</table>

app/src/main/java/com/google/maps/android/compose/MarkerClusteringActivity.kt
```diff
@@ -107,6 +108,181 @@ fun GoogleMapClustering(items: List<MyItem>) {
...
+        }
+        clusterManager.setOnClusterItemInfoWindowClickListener {
+            Log.d(TAG, "Cluster item info window clicked! $it")
+        }
+    }
+    SideEffect {
+        if (clusterManager?.renderer != renderer) {
+            clusterManager?.renderer = renderer ?: return@SideEffect
+        }
+    }
```
matsudamper, line 210: For the high-level APIs, Processing is performed in Clustering Composable.
For the low-level API, I wondered what to do, but did not include it. Because I may be some inconvenience I'm not aware of.

matsudamper:
> java.lang.IllegalStateException: Maps API key not specified
https://github.com/googlemaps/android-maps-compose/actions/runs/6847027102/job/18629617389?pr=421

The test was successful in my environment. I didn't know the cause.

wangela:
This will also help us customize the algorithm for developing and testing fixes to #380

wangela (approved):
Thank you for contributing this feature, @matsudamper!

Issue #339, opened 2023-07-21
Make zooming controls customizable


vladcioricagoogle:
Current zooming controls are not customizable at all and also they are not supporting dark mode.

wangela:
If you would like to upvote the priority of this issue, please comment below or react with :+1: so we can see what is popular when we triage.
--------

@vladcioricagoogle Thank you for opening this issue. 🙏
Please check out these other resources that might help you get to a resolution in the meantime:

- [Check the issue tracker](https://developers.google.com/maps/support/#issue_tracker) - bugs and feature requests for Google Maps Platform APIs and SDKs
- [Open a support case](https://developers.google.com/maps/support#contact-maps-support) - Get 1:1 support in Cloud Console.
- [Discord](https://discord.gg/fHTDmZe9En) - chat with other developers
- [StackOverflow](http://stackoverflow.com/questions/tagged/google-maps) - use the `google-maps` tag


_This is an automated message, feel free to ignore._

wangela:
Once the pattern is established for the My Location button in #347 you can implement this for the zoom controls or other map controls as needed, @vladcioricagoogle.

vladcioricagoogle:
@kikoso after merging the PR for the location button is it possible to take care of this because you already have enough context on the subject and it should be pretty similar? Thanks!

vladcioricagoogle:
@kikoso LGTM, thank you very much!

PR #428, opened 2023-10-11
feat!: setting the custom location button via sample

### LINKED ISSUE #427: v3.1.0 seems to ignore Modifier.weight(1f) in Column
I included a GoogleMap inside a Column along with a dropdown at the top. Something like

```kotlin
Column(
    modifier = Modifier.fillMaxSize() // entire screen
) {
    MapTypeSelector(
        currentValue = currentMapType,
        modifier = Modifier.fillMaxWidth(),
    )
    GoogleMap(
        modifier = Modifier
            .fillMaxWidth()
            .weight(1f),
    )
}
```

With 3.0.0 I see what I expect:

With 3.1.0 the same code displays the map on only the top half of the screen:

kikoso:
Update: this PR moves the custom location to the sample instead of managing it via the GoogleMap.

~~The following PR fixes an issue when a Modifier was set for GoogleMaps. This PR introduces a custom Modifier for the custom location row in the map, and adds this one only when a custom location has been added.~~

---

Fixes #427

### DIFF
--- a/maps-compose/src/main/java/com/google/maps/android/compose/GoogleMap.kt
+++ b/maps-compose/src/main/java/com/google/maps/android/compose/GoogleMap.kt
@@ -74,6 +74,7 @@ import kotlinx.coroutines.awaitCancellation
 @Composable
 public fun GoogleMap(
     modifier: Modifier = Modifier,
+    currentLocationModifier : Modifier = Modifier,
     cameraPositionState: CameraPositionState = rememberCameraPositionState(),
     contentDescription: String? = null,
     googleMapOptionsFactory: () -> GoogleMapOptions = { GoogleMapOptions() },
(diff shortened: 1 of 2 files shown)


maps-compose/src/main/java/com/google/maps/android/compose/GoogleMap.kt
```diff
@@ -74,6 +74,7 @@ import kotlinx.coroutines.awaitCancellation
 @Composable
 public fun GoogleMap(
     modifier: Modifier = Modifier,
+    currentLocationModifier : Modifier = Modifier,
```
bubenheimer, line 77: Having two Modifiers is hard to understand to a user and makes the already rich GoogleMap() API surface even heavier.

This is needed only because of the changes introduced in #347. I pointed out several problems in https://github.com/googlemaps/android-maps-compose/issues/340#issuecomment-1753175867 and https://github.com/googlemaps/android-maps-compose/issues/340#issuecomment-1753252034.

My suggestion is to move the changes from #347 into an official example instead of including them in the API to address these problems and to reduce API bloat.

I understand #347 is your baby, but please take a step back and reconsider the options. Adding custom buttons in the fashion of #347 makes for a poor API experience.

bubenheimer (approved):

bubenheimer (approved):

bubenheimer (approved):

wangela:
Decisions we make here will also steer the implementation approach for #339.

vladcioricagoogle:
@kikoso looks good to me, seems that we still can achieve the required functionality so I don't have anything agains it

maps-compose/src/main/java/com/google/maps/android/compose/GoogleMap.kt
wangela: Thank you for explaining the rationale, @bubenheimer. We are merging this now as a breaking change.

wangela (approved):

Issue #403, opened 2023-09-21
[Support] Hide marker on google map compose


GDeath019:
Hello
I am a dev in Viet Nam. Now I am doing with google map in compose and I have a issue.
When map showing I want hide all marker (from default of google: maybe like restaurant, market, ...) but I can't find a function or setting can hidden marker.
So please help us create a feature for that or give me a guide for this issue.
Thank you

wangela:
If you would like to upvote the priority of this issue, please comment below or react on the original post above with :+1: so we can see what is popular when we triage.
--------

@GDeath019 Thank you for opening this issue. 🙏
Please check out these other resources that might help you get to a resolution in the meantime:

- [Check the issue tracker](https://developers.google.com/maps/support/#issue_tracker) - bugs and feature requests for Google Maps Platform APIs and SDKs
- [Open a support case](https://developers.google.com/maps/support#contact-maps-support) - Get 1:1 support in Cloud Console.
- [Discord](https://discord.gg/fHTDmZe9En) - chat with other developers
- [StackOverflow](http://stackoverflow.com/questions/tagged/google-maps) - use the `google-maps` tag


_This is an automated message, feel free to ignore._

binayshaw7777:
I guess you can achieve this using `MapEffect() {...}` inside `GoogleMap(){...}` scope. 
Map effect passes a lambda, `map ->` which can be used to monitor the zoom level like -> ```map.setOnCameraMoveListener {
                                val currentZoomLevel = map.cameraPosition.zoom
                                }
                                ```
                                
Similarly, you can use other methods as well and add your markers accordingly. Check this out [Google Map - Public Method Summary](https://developers.google.com/android/reference/com/google/android/gms/maps/GoogleMap#public-method-summary)

I hope this helps you somehow or least gives some lead.

Issue #91, opened 2022-04-04
How to clear map markers


tnt-ndricimh:

jpoehnelt:
@tnt-ndricimh Thank you for opening this issue. 🙏
Please check out these other resources that might be applicable:

- [Support Console](https://cloud.google.com/support/) - If you have a support contract.
- [Discord](https://discord.gg/9UpV862xZ9) - chat with other developers
- [StackOverflow](http://stackoverflow.com/questions/tagged/google-maps) - use the `google-maps` tag
- [Google Maps Platform issue trackers](https://developers.google.com/maps/support/#issue_tracker)


_This is an automated message, feel free to ignore._

barbeau:
@tnt-ndricimh Have you tried removing all the content from the composable when you want an empty map? Normally with composables it shows whatever you provide as content. For example, the boolean value `showMarker` controls whether or now the Marker is visible below. If you set this to `false` then the marker should disappear.

```kotlin
...
val showMarker = remember { mutableStateOf(true) }

GoogleMap(
    ...
    ) {
        // Marker is only shown if showMarker is true
       if (showMarker) {
            Marker(
                state = singapore3State,
                title = "Marker in Singapore",
            )
       }
    }
```

rtippise:
I guess it works with single or couple of markers.
I have a use case where I would like to show nearby 100+ locations as markers on the map. When I drag the map to a new location I would like to show new 100+ nearby locations on the map. During the process, I need to remove the old 100 markers and place new 100+ markers (number varies). I am wondering how can we achieve this dynamic loading of markers with the Compose library.
`
val locations by locationsStateFlow.collectAsState() 
GoogleMap(
                 ...
  ) {
       for(location in locations) {
              Marker(
                state = singapore3State,
                title = "Marker in Singapore",
            )      
        }
}`

arriolac:
So long as the value in your for loop contains the new list of locations you'd like to display markers for (not additive to the previous list), the code as you have written will ensure that only the new list of markers will be displayed (i.e. the previous markers should be removed).

barbeau:
Something to watch out for is that you'll need to make sure that your Flow is returning an immutable collection of locations that changes to a new collection instance every time. 

https://developer.android.com/jetpack/compose/state#use-other-types-of-state-in-jetpack-compose says:

>Caution: Using mutable objects such as ArrayList<T> or mutableListOf() as state in Compose will cause your users to see incorrect or stale data in your app.
>
[...]

PR #448, opened 2023-10-27
docs: added recomposition sample

### LINKED ISSUE #442: Duplication of marker during update of viewmodel
Hello,

I am using the Google Maps Android Maps Compose in my project and I've encountered an issue. When I update (recompose) the marker state from my ViewModel, it results in the duplication of the existing marker.

Steps to reproduce:
1. Set up a ViewModel that listens to changes from a database.
2. In the ViewModel, create a list of `Places` containing origin and destination `latlng`.
3. Iterate over the list of `Places`, creating a pair of markers one for origin one for destination for each place.
4. Update (recompose) the marker state from the ViewModel.
```kt
viewModel.fetchData.toMutableStateList().filterNotNull().forEach { dataItem ->

    cameraPositionState.position = CameraPosition.fromLatLngZoom(dataItem.location.currentPosition.latLng(), 17f)
    val currentMarkerState = MarkerState(dataItem.location.currentPosition.latLng())
    val destinationMarkerState = MarkerState(dataItem.location.destinationPosition.latLng())

    Marker(
        state = currentMarkerState,
        ...
    )
    Marker(
        state = destinationMarkerState,
      ...
    )
}

```
Expected behavior:
I expected the marker state to update without duplicating the existing marker.

Actual behavior:
The existing marker is duplicated when the marker state is updated from the ViewModel.

I would appreciate any guidance on this issue. Thank you for your time and consideration.

kikoso:
This PR adds a recomposition sample, to showcase how to remove or add markers using a remember variable and recomposition.

https://github.com/googlemaps/android-maps-compose/assets/903097/098a6a95-4a24-45e0-b5e8-1acfa3b144ce

---

Fixes #442, #91, #403 🦕

### DIFF
--- a/README.md
+++ b/README.md
@@ -201,6 +201,30 @@ GoogleMap(
 </details>
 
 <details>
+  <summary>Recomposing elements</summary>
+
+### Recomposing elements
+
+Markers and other elements need to be recomposed in the screen. To achieve recomposition you need to 
+have a mutable state variable:
+
+```kotlin
+var location by remember { mutableStateOf(singapore) }
+val singaporeState = MarkerState(position = location)
+
+//...
+
+Marker(
+    state = singaporeState,
+    title = "Marker in Singapore",
+    onClick = markerClick
+)
+```
+
+In the example above, when `location` changes the recomposition will be triggered. 
+
+</details>
+<details>
   <summary>Customizing a marker's info window</summary>
 
 ### Customizing a marker's info window
--- a/app/src/main/java/com/google/maps/android/compose/RecompositionActivity.kt
+++ b/app/src/main/java/com/google/maps/android/compose/RecompositionActivity.kt
@@ -1,0 +1,118 @@
+// Copyright 2023 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package com.google.maps.android.compose
+
+import android.os.Bundle
+import android.util.Log
+import androidx.activity.ComponentActivity
+import androidx.activity.compose.setContent
+import androidx.compose.foundation.layout.Box
+import androidx.compose.foundation.layout.Column
+import androidx.compose.foundation.layout.fillMaxSize
+import androidx.compose.material.Button
+import androidx.compose.material.Text
+import androidx.compose.runtime.Composable
+import androidx.compose.runtime.getValue
+import androidx.compose.runtime.mutableStateOf
+import androidx.compose.runtime.remember
+import androidx.compose.runtime.setValue
+import androidx.compose.ui.Modifier
+import com.google.android.gms.maps.model.Marker
+import com.google.maps.android.compose.theme.MapsComposeSampleTheme
+import kotlin.random.Random
+
+private const val TAG = "RecompositionActivity"
+
+/**
+ * This is a sample activity showcasing how the recomposition works. The location is changed
+ * every time we click on the button, and the marker gets updated (removed and added in a new
+ * location)
+ */
+class RecompositionActivity : ComponentActivity() {
+
+    override fun onCreate(savedInstanceState: Bundle?) {
+        super.onCreate(savedInstanceState)
+        setContent {
+            val cameraPositionState = rememberCameraPositionState {
+                position = defaultCameraPosition
+            }
+            Box(Modifier.fillMaxSize()) {
+                MapsComposeSampleTheme {
+                    GoogleMapView(
+                        modifier = Modifier.matchParentSize(),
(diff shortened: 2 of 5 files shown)


wangela (changes requested):
Let's also add to the "Drawing on a map" section of the README about how to update the state of composables that have been added to the map. Can just link to this activity if the code snippet gets too long.

wangela (approved):

bubenheimer:
@wangela May I ask why these commits were merged without further discussion? I pointed out a problem, yet my comments were ignored. This action does not promote a healthy open-source culture for this project. I will need to assess the relevance of future contributions from my side.

I respect that there may be differences in opinion, but there needs to be discussion to resolve those. Apart from my prior pointers: the API provides a dedicated function [`rememberMarkerState()`](https://googlemaps.github.io/android-maps-compose/maps-compose/com.google.maps.android.compose/remember-marker-state.html) to maintain `MarkerState`. It is unclear why the project documentation and samples would promote a different approach that does not actually remember the state object itself, but continuously recreates it.

I tried to voice my concerns in a helpful, professional manner, but perhaps I was too subtle? I would not go through the effort of commenting unless I see a strong need for changes.

The present action suggests that I and other users of this project will need to find alternative means to restore and maintain sufficient API quality for our own projects in the face of ongoing regressions since @arriolac's departure. For example the present documentation regression, as well as #466 and #347. I appreciate that #347 was reverted after I voiced my concerns, but this PR points in a different direction. The existence of #466 should have rung alarm bells.

The end result of all this is a necessity for private forks. I do not see a viable alternative, unless there are changes to how this project is managed. Seeing #347 gave me a scare (its problems should have been caught in code reviews, and it even required changing a test to make it pass), so I started to review every PR personally in an effort to catch serious regressions early, but now I am finding that my input is ignored. Private forks are expensive to maintain for independent devs like myself, so I wish this could be resolved in a different manner.

bubenheimer:
@kikoso I'm not sure what you mean about "no comment" on this PR. I started a code review right here in this PR soon after your initial commit, with 4 comments total. You did not respond to the concerns that I brought up in the context of the review. It's all still up there. Please clarify.

bubenheimer:
As for whether to use `rememberMarkerState()` specifically: it's not essential, it just `remembers` 'MarkerState'.

What I'm saying is that it does not make sense to recreate `MarkerState` on recomposition, and `remember` a separate `mutableStateOf()` for the location. `MarkerState` is the relevant state object here and needs to be the one to be remembered. `MarkerState` encapsulates more than just location.

The merged commit demonstrates a bad pattern. I hope it can be unmerged.

app/src/main/java/com/google/maps/android/compose/RecompositionActivity.kt
```diff
@@ -0,0 +1,118 @@
...
+    }
+
+    @Composable
+    fun GoogleMapView(
+        modifier: Modifier = Modifier,
+        cameraPositionState: CameraPositionState = rememberCameraPositionState(),
+        content: @Composable () -> Unit = {},
+    ) {
+        var location by remember { mutableStateOf(singapore) }
+        val singaporeState = MarkerState(position = location)
```
bubenheimer, line 69: `MarkerState.position` is a var with a public setter and a `mutableStateOf` behind it, and it is observed for changes by `Marker()` machinery. So it is not necessary to keep a separate remembered `mutableStateOf` to track position changes, or to recreate MarkerState on every recomposition.
Consider this instead:
```
val singaporeState = remember { MarkerState(singapore) }
```
Then update via:

```
singaporeState.position = when (randomValue) {
    0 -> singapore
    1 -> singapore2
    2 -> singapore3
    else -> singapore
```

bubenheimer: @wangela @kikoso if this pattern does not work then there has been a regression that needs to be fixed. Various recently filed issues and my own testing suggest that this is the case. The pattern used to work flawlessly and is the way the API is intended to be used; it's described at https://github.com/androidx/androidx/blob/androidx-main/compose/docs/compose-api-guidelines.md#hoisted-state-types and subsequent sections.

Recreating MarkerState on every recomposition is akin to the following, which the IDE will tell you is missing 'remember':
```
val state = mutableStateOf(1)
```

bubenheimer: @wangela @kikoso FYI: the regression (if it is just a single one) was introduced in 2.14.1

README.md
```diff
@@ -200,6 +200,30 @@ GoogleMap(
...
+  <summary>Recomposing elements</summary>
+
+### Recomposing elements
+
+Markers and other elements need to be recomposed in the screen. To achieve recomposition you need to 
+have a mutable state variable:
+
+```kotlin
+var location by remember { mutableStateOf(singapore) }
+val singaporeState = MarkerState(position = location)
```
bubenheimer, line 213: Same concerns as in sample

bubenheimer:
Thanks @kikoso, I must have missed clicking the button initially, and I think the button may have stopped working later after force push & merge/closure. I've added my prior comments as regular comments now, to show what I had in mind. I seem no longer able to start a review on a closed PR.

My sincere apologies for the noise & waste of time on your end.

Would it be possible to change the sample & docs still? I don't think there is any valid use case where one would not want to retain MarkerState across recompositions.

That said, there certainly is a common need to sync MarkerState with information that is kept outside Compose. This would be a slightly larger example. An important example, though.

PR #473, opened 2023-12-12
Allow `minClusterSize` to pass-through on `Clustering`

### LINKED ISSUE #388: Allow a way to set options for Clustering like setMinClusterSize
I'm really missing a way to customize the MinClusterSize on the Clustering composable.

darronschall:
When using the default cluster renderer, allow callers to easily configure the `minClusterSize` value.

Thank you for opening a Pull Request!

---

Fixes #388 🦕


Jasperav:
I actually think it's easier to add a lambda to the current method which can look like this:

`modifyDefaultClusterRenderer: (DefaultClusterRenderer) -> Unit = { }`

and than users can just do whatever they want with it, so you add 1 method to the existing method and call the lambda after creating the default cluster renderer.

darronschall:
It's been over 6 months since I initially created this PR. I noticed recently it was stale and needed conflict resolution.

I rebased and rebuilt it to simplify history. Any chance this version can get merged? It closes #388 with a minimal optional API addition.

PR #478, opened 2023-12-16
fix: Use recomposed map listener callbacks rather than only the initially …

### LINKED ISSUE #326: Content padding is not used for initial rendering

### LINKED ISSUE #466: Regression: recomposed map listeners ignored
A regression was introduced in 1.4.21, specifically #320, where only the initial compositions of all the GoogleMap listener lambdas are used; recomposed listener lambdas are ignored. This can cause pervasive malfunctions of listener functionality and is a likely cause for some of the issues reported in the last months.

CodeReview

The diff screenshot shows the culprit - the actual Map listener used to call the latest version of the listener lambda, whereas now the actual Map listener is directly set to the initial version of the listener lambda. 

I don't know how to trivially fix it without undoing this part of the PR, which would undo the fix for #38. @romainpiel, as the original author of the PR, are you able to devise a way to get recomposed listeners working again?

I am seeing this issue as of maps-compose version 4.3.0, but it has been around since 1.4.21.

bubenheimer:
…composed version

Fixes #466

This solution is optimized for the common case of recomposition updating a non-null GoogleMap listener callback to another non-null callback, and has negligible overhead, supporting rapid recomposition.

The other, less common case, toggling between null and non-null callbacks, calls the appropriate GoogleMap SDK listener setter in the composition apply phase; any overhead from calling the setter is passed through.

Thank you for opening a Pull Request!

---

### DIFF
--- a/maps-compose/src/main/java/com/google/maps/android/compose/GoogleMap.kt
+++ b/maps-compose/src/main/java/com/google/maps/android/compose/GoogleMap.kt
@@ -27,7 +27,6 @@ import androidx.compose.runtime.CompositionLocalProvider
 import androidx.compose.runtime.DisposableEffect
 import androidx.compose.runtime.LaunchedEffect
 import androidx.compose.runtime.MutableState
-import androidx.compose.runtime.SideEffect
 import androidx.compose.runtime.getValue
 import androidx.compose.runtime.mutableStateOf
 import androidx.compose.runtime.remember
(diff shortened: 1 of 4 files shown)


bubenheimer:
@kikoso would you have time for a review? Happy to explain changes. Would like to get it merged soon to not let the issue fester longer. Have a few other PRs waiting on this one.

bubenheimer:
Bump

bubenheimer:
I've reverted the earlier commit that moved updating the listeners into a SideEffect. The SideEffect approach delays the listener updates to a future recomposition cycle, which I find too slow.

Without SideEffect there is the issue of spurious recompositions that I believe are fast and do essentially nothing. However, the listeners are updated more quickly. I have not seen issues in terms of affected code behaving incorrectly because of the spurious recompositions.

The spurious recompositions are not specific to this PR, they are prone to happen with the current GoogleMap() subcomposition approach in general: #480

PR #514, opened 2024-01-28
fix: make DefaultMapProperties, DefaultMapUiSettings, and NoPadding public

### LINKED ISSUE #513: Default expressions on optional parameters are not public
android-maps-compose 4.3.2

The `GoogleMap()` compose function uses certain default expressions on optional parameters that are not public. This prevents  overriding the function in app code in a reasonable manner.

This is what the "API Guidelines for @Composable components in Jetpack Compose" have to say on it:

> Developers should make sure that default expressions on optional parameters are publicly available and meaningful.

https://github.com/androidx/androidx/blob/androidx-main/compose/docs/compose-component-api-guidelines.md#default-expressions

bubenheimer:
Fixes #513


bubenheimer:
Thanks, @kikoso. I added documentation and renamed nondescript top level `NoPadding` to `DefaultMapContentPadding`.

PR #515, opened 2024-01-29
fix: improve MarkerState API

### LINKED ISSUE #149: Marker dragging: fix loss of drag events and difficulty observing marker positions

bubenheimer:
This is a non-breaking change following suggestions from @arriolac for addressing #149: https://github.com/googlemaps/android-maps-compose/pull/150#discussion_r1016963262

This PR does not add an `onDrag` callback parameter to `Marker()`, which would be a somewhat breaking change; this functionality is not strictly necessary and I see alternatives that may be preferable.

Summary of changes:

1. Deprecate MarkerState.dragState and DragState enum. These were carried over from GoogleMap SDK; they are events that were mischaracterized as states.
2. Replace with MarkerState.isDragging boolean.
3. Clarify KDoc in several places.
4. Add several examples providing patterns for common use cases.
5. Organize Marker-related examples into their own folder.

Fixes #149

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -43,13 +43,13 @@
         android:name=".BasicMapActivity"
         android:exported="false" />
     <activity
-        android:name=".marker.AdvancedMarkersActivity"
+        android:name=".markerexamples.AdvancedMarkersActivity"
         android:exported="false"/>
     <activity
         android:name=".MapInColumnActivity"
         android:exported="false"/>
     <activity
-        android:name=".marker.MarkerClusteringActivity"
+        android:name=".markerexamples.MarkerClusteringActivity"
         android:exported="false"/>
     <activity
         android:name=".LocationTrackingActivity"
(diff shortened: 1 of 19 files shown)


bubenheimer:
To be clear: replacing `MarkerState.dragState` (START/DRAG/END) with `MarkerState.isDragging` does not lose information, because the information was not valid in the first place, due to the State vs. Event mismatch in `dragState`.

The MarkerDragEventsActivity example doubles as a proof of concept demonstrating how to recreate `DragState` events from `MarkerState.isDragging` plus `MarkerState.position`.

bubenheimer:
@wangela can I get a review for this PR, please? I'd really like to use these improvements. Thanks.

dkhawk (approved):

dkhawk (approved):

PR #517, opened 2024-01-29
fix: add stability configuration file to the core maps-compose projec…

### LINKED ISSUE #152: Performance penalty throughout API from LatLng usages
android-maps-compose 2.2.1
compose 1.2.0-rc01

The android-maps-compose API imposes a performance penalty throughout by leveraging the Maps SDK LatLng type, which the compiler is currently unable to infer as stable. (Observed via compose compiler metrics: https://github.com/androidx/androidx/blob/androidx-main/compose/compiler/design/compiler-metrics.md)

I see 2 potential solutions:

1. Replace LatLng with an equivalent stable type throughout maps-compose API.
2. Work with compose compiler team to treat LatLng as stable.

I'd consider the second solution preferable.
#### Code example

    @Composable
    fun MyComposable(latLng: LatLng) {}

In this example the Compose compiler deems `latLng` unstable, rendering the composable function non-skippable.

bubenheimer:
…t. The currently configured types are all the immutable data types from com.google.android.gms.maps.model that are currently used in android-maps-compose:maps-compose. These types will now be considered @Stable by the compiler. Making immutable data types @Stable can be a more effective measure than relying on strong skipping, as stable types can use structural equality, whereas strong skipping merely relies on referential equality, as of today.

This commit also makes compose compiler report generation available via command-line options (-PcomposeCompilerMetrics=true, -PcomposeCompilerReports=true)

Fixes #152

### DIFF
--- a/maps-compose/build.gradle
+++ b/maps-compose/build.gradle
@@ -31,6 +31,27 @@ android {
         jvmTarget = '1.8'
         freeCompilerArgs += '-Xexplicit-api=strict'
         freeCompilerArgs += '-Xopt-in=kotlin.RequiresOptIn'
+
+        freeCompilerArgs += [
+                "-P",
+                "plugin:androidx.compose.compiler.plugins.kotlin:stabilityConfigurationPath=" +
+                        layout.projectDirectory.file('compose_compiler.conf').asFile.absolutePath
+        ]
+
+        if (findProperty("composeCompilerReports") == "true") {
+            freeCompilerArgs += [
+                    "-P",
+                    "plugin:androidx.compose.compiler.plugins.kotlin:reportsDestination=" +
+                            project.buildDir.absolutePath + "/compose_compiler"
+            ]
+        }
+        if (findProperty("composeCompilerMetrics") == "true") {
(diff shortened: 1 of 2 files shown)


bubenheimer:
@kikoso are you open to reviewing this PR?

bubenheimer:
@wangela can this PR receive a review, please?

bubenheimer:
@kikoso The types in `model` are not generally immutable, which commonly would make them non-stable for Compose purposes. They would be covered by the upcoming strong skipping changes for handling non-stable types in Compose, which is a little less powerful than what we can achieve with stability configuration files.

I've validated for the types in the stability configuration file that they are immutable. These are all the Maps SDK model types currently in use in the maps-compose library.

I did not look at what's being used in the maps-compose-utils, and did not provide a stability configuration file there, as I don't use that lib, not familiar with it. It can be done independently of this PR.

bubenheimer:
@kikoso Actually, `maps-compose` uses some mutable types from `model` as well (e.g. Marker), but they need to be handled via strong skipping changes, not stability configuration file, for the the reason I gave above.

PR #533, opened 2024-03-15
fix: Return early in rememberComposeBitmapDescriptor for invalid view size

### LINKED ISSUE #518: MarkerComposable throw `Fatal Exception: java.lang.IllegalArgumentException: width and height must be > 0`
Hello, 
I'm working with Google Maps Compose and now I get some Firebase crashes related to markerComposable  withe the following exception:

```
Fatal Exception: java.lang.IllegalArgumentException: width and height must be > 0
       at android.graphics.Bitmap.createBitmap(Bitmap.java:1148)
       at android.graphics.Bitmap.createBitmap(Bitmap.java:1115)
       at android.graphics.Bitmap.createBitmap(Bitmap.java:1065)
       at android.graphics.Bitmap.createBitmap(Bitmap.java:1026)
       at com.google.maps.android.compose.RememberComposeBitmapDescriptorKt.renderComposableToBitmapDescriptor(RememberComposeBitmapDescriptor.kt:58)
       at com.google.maps.android.compose.RememberComposeBitmapDescriptorKt.access$renderComposableToBitmapDescriptor(RememberComposeBitmapDescriptor.kt:1)
       at com.google.maps.android.compose.RememberComposeBitmapDescriptorKt.rememberComposeBitmapDescriptor(RememberComposeBitmapDescriptor.kt:29)
       at com.google.maps.android.compose.MarkerKt.MarkerComposable-pMp3byo(Marker.kt:239)
       ```

this is my following code : 

```
    @Composable
    fun MapScreen(
     ...
    ) {

     GoogleMap(
           modifier = Modifier
              .fillMaxWidth()
              .padding(bottom = sheetPeekHeight - 8.dp),
          properties = if (locationPermissionState.status.isGranted) MapProperties(isMyLocationEnabled = true) else MapProperties(
              isMyLocationEnabled = false
         ),
          uiSettings = MapUiSettings(zoomControlsEnabled = false)
       ) {
          LocationMarkers(locations, getMarkerSequence)
     } 
    }

    @Composable
    fun LocationMarkers(
       locations: List<LatLng>,
       getMarkerSequence: (Int) -> String
     ) {

      locations.onEachIndexed { index, latLng ->

        MarkerComposable(
            state = MarkerState(latLng),
        ) {
            val markerSequence = getMarkerSequence(index)
            Box(contentAlignment = Alignment.Center) {
                Image(
                    painter = painterResource(id = R.drawable.ic_my_icon),
                    contentDescription = ""
                )

                Text(
                    text = markerSequence,
                    color = Color.White,
                    modifier = Modifier
                        .padding(top = 4.dp)
                        .align(Alignment.TopCenter)
                )
            }

        }
      }
     }

#### Environment details

1. API  Google Maps compose 
2. it's happening at multiple Version From Android 9 ->14
3. Google Maps compose version [v2.14.0](https://github.com/googlemaps/android-maps-compose/releases/tag/v2.14.0)

#### Steps to reproduce

  1.  I can't reproduce my local but the crashes caught by Firebase affect around 1% of the users


Thanks!

ln-12:
Fixes #518  🦕

By returning early in the case of invalid view width/height, we can avoid the IllegalArgumentException described in issue #518.

### DIFF
--- a/maps-compose/src/main/java/com/google/maps/android/compose/RememberComposeBitmapDescriptor.kt
+++ b/maps-compose/src/main/java/com/google/maps/android/compose/RememberComposeBitmapDescriptor.kt
@@ -30,6 +30,8 @@ internal fun rememberComposeBitmapDescriptor(
     }
 }
 
+private val measureSpec = View.MeasureSpec.makeMeasureSpec(0, View.MeasureSpec.UNSPECIFIED)
+
 private fun renderComposableToBitmapDescriptor(
     parent: ViewGroup,
     compositionContext: CompositionContext,
@@ -50,10 +52,7 @@ private fun renderComposableToBitmapDescriptor(
 
     composeView.draw(fakeCanvas)
 
-    composeView.measure(
-        View.MeasureSpec.makeMeasureSpec(0, View.MeasureSpec.UNSPECIFIED),
-        View.MeasureSpec.makeMeasureSpec(0, View.MeasureSpec.UNSPECIFIED),
-    )
+    composeView.measure(measureSpec, measureSpec)
 
     if (composeView.measuredWidth == 0 || composeView.measuredHeight == 0) {
         throw IllegalStateException("The ComposeView was measured to have a width or height of " +
--- a/maps-compose/src/main/java/com/google/maps/android/compose/RememberComposeBitmapDescriptor.kt
+++ b/maps-compose/src/main/java/com/google/maps/android/compose/RememberComposeBitmapDescriptor.kt
@@ -56,7 +56,7 @@ private fun renderComposableToBitmapDescriptor(
 
     if (composeView.measuredWidth == 0 || composeView.measuredHeight == 0) {
         throw IllegalStateException("The ComposeView was measured to have a width or height of " +
-                "zero. Make sure the parent and content have a non-zero size.")
+                "zero. Make sure that the content has a non-zero size.")
     }
 
     composeView.layout(0, 0, composeView.measuredWidth, composeView.measuredHeight)
--- a/maps-compose/src/main/java/com/google/maps/android/compose/RememberComposeBitmapDescriptor.kt
+++ b/maps-compose/src/main/java/com/google/maps/android/compose/RememberComposeBitmapDescriptor.kt
@@ -39,6 +39,10 @@ private fun renderComposableToBitmapDescriptor(
     val composeView =
         ComposeView(parent.context)
             .apply {
+                layoutParams = ViewGroup.LayoutParams(
+                    ViewGroup.LayoutParams.WRAP_CONTENT,
+                    ViewGroup.LayoutParams.WRAP_CONTENT,
+                )
                 setParentCompositionContext(compositionContext)
                 setContent(content)
             }
@@ -47,9 +51,14 @@ private fun renderComposableToBitmapDescriptor(
     composeView.draw(fakeCanvas)
 
     composeView.measure(
-        View.MeasureSpec.makeMeasureSpec(parent.width, View.MeasureSpec.AT_MOST),
-        View.MeasureSpec.makeMeasureSpec(parent.height, View.MeasureSpec.AT_MOST),
+        View.MeasureSpec.makeMeasureSpec(0, View.MeasureSpec.UNSPECIFIED),
+        View.MeasureSpec.makeMeasureSpec(0, View.MeasureSpec.UNSPECIFIED),
     )
+
+    if (composeView.measuredWidth == 0 || composeView.measuredHeight == 0) {
+        throw IllegalStateException("The ComposeView was measured to have a width or height of " +
+                "zero. Make sure the parent and content have a non-zero size.")
+    }
 
     composeView.layout(0, 0, composeView.measuredWidth, composeView.measuredHeight)
 
--- a/maps-compose/src/main/java/com/google/maps/android/compose/Marker.kt
+++ b/maps-compose/src/main/java/com/google/maps/android/compose/Marker.kt
@@ -200,6 +200,8 @@ public fun Marker(
 /**
  * Composable rendering the content passed as a marker.
  *
+ * This composable must have a non-zero size in both dimensions
+ *
  * @param keys unique keys representing the state of this Marker. Any changes to one of the key will
  * trigger a rendering of the content composable and thus the rendering of an updated marker.
  * @param state the [MarkerState] to be used to control or observe the marker
@@ -221,6 +223,9 @@ public fun Marker(
  * @param onInfoWindowClose a lambda invoked when the marker's info window is closed
  * @param onInfoWindowLongClick a lambda invoked when the marker's info window is long clicked
  * @param content composable lambda expression used to customize the marker's content
+ *
+ * @throws IllegalStateException if the composable is measured to have a size of zero in either
+ * dimension
  */
 @Composable
 @GoogleMapComposable


maps-compose/src/main/java/com/google/maps/android/compose/RememberComposeBitmapDescriptor.kt
```diff
@@ -51,6 +51,11 @@ private fun renderComposableToBitmapDescriptor(
         View.MeasureSpec.makeMeasureSpec(parent.height, View.MeasureSpec.AT_MOST),
     )
 
+    if(composeView.measuredWidth == 0 || composeView.measuredHeight == 0) {
+        parent.removeView(composeView)
```
bubenheimer, line 55: You can avoid the extra removeView() by adding try-finally in the right place here.

bubenheimer (commented):
The end result of this change is that the map will render the default marker instead of crashing on an empty composable marker. This would seem to hide the problem in a workaround, rather than flag it. Would throwing a more meaningful exception help with troubleshooting here?

Just my 2 cents here: creating an empty ComposeView suggests some kind of problem in user code. I imagine user code may be able to avoid creating the marker entirely by testing whether necessary preconditions are met before attempting to create the marker.

maps-compose/src/main/java/com/google/maps/android/compose/RememberComposeBitmapDescriptor.kt
ln-12: Thanks for the feedback! Yeah, unfortunately I am not sure why this happens in the first place. Maybe some race condition while navigating? I can't dig into it as this happens super rarely and I was not able to reproduce it while having the debugger attached.

I don't think that a different error would help here as we already know that the problem is an incorrect view size. The question is just why we get to this state. As you can see from [the example code](https://github.com/googlemaps/android-maps-compose/issues/518#issue-2105158020) in the issue, the original reporter only uses the GoogleMaps composable and adds some markers (the same as I do in my code). I don't see what we as users of this library could do differently as the code works >99% of the time. Or could we add any meaningful information to a new exception which would help finding the root cause?

bubenheimer: What if you used `Modifier.requiredSizeIn()` on your marker composable content with a minimum of 1.dp or so, as an alternative workaround? I'd be curious if you still crash then.

bubenheimer: @ln-12 Actually, I suspect that setting a minimum size on the composable, like I suggested above, would not make a difference.

Here is the problem that I think may be at the root of this issue (have not tested it):
https://github.com/googlemaps/android-maps-compose/blob/0599412f11fb0571c97a52d0b29a4943f2a60a65/maps-compose/src/main/java/com/google/maps/android/compose/RememberComposeBitmapDescriptor.kt#L50-L51

A parent view dimension of 0 would cause this crash, no matter how large the Composable wants to be.

I'd think the parent view here would commonly be the Activity's top ComposeView, or supposedly something else in case of navigation, etc. So the maximum marker icon size would commonly be constrained by the size of the screen, which looks weird to me, as that could be quite large.

A better approach, short of workarounds, may be to use `View.MeasureSpec.UNSPECIFIED` here; restricting marker size by screen size just does not look that useful, it's way too big, creating giant bitmaps. However, I imagine this change would be breaking, as there is no max size anymore. Yet a marker composable relying on something like fillMaxSize is not doing it right, it needs to use something more specific. There may be other concerns with UNSPECIFIED, however, not sure.

ln-12: You are right. I can force the crash by setting the size of the fragment which contains the compose view to 0 (but I am still not sure when this happens while navigating): https://github.com/ln-12/GoogleMapsCrashReproducer/blob/main/app/src/main/res/layout/activity_main.xml#L11-L12

With that, I also tested `View.MeasureSpec.UNSPECIFIED` and for me it did not crash anymore. I can change my PR to use this spec, but just like you, I am not sure about other side effects when doing so.

So what would you suggest going forward?

bubenheimer: Thank you for these useful observations.

To me `MeasureSpec.UNSPECIFIED` feels like the right thing to do here, as there is no valid relationship between marker size and screen size (assuming marker size is expressed via dips), so relying on parent size for calculating marker size is invalid. `MeasureSpec.AT_MOST` seems to trade more predictable crashes (from potentially relying on absent parent size) for random crashes.

Also, as we see in this case, parent size can change over time, but this would not trigger recalculation of the marker based on the new size.

Not sure if the ComposeView should also set layoutParams, using `LayoutParams.WRAP_CONTENT`. Could some layouts choke on WRAP_CONTENT?

I see that @dkhawk is assigned to this PR, he may need to take over from here.

dkhawk: @bubenheimer Thanks for being this to my attention.  I would love talk to you more about it and your many other contributions.  Please reach out to me via my Twitter (@DaleHawkin57064) if that works for you.  Or via LinkedIn: https://www.linkedin.com/in/dalehawkins/.

bubenheimer: If no one objects, I suggest going with the following (see above comment):
1. Set layoutParams (WRAP_CONTENT) on ComposeView, because docs say you have to.
2. Use View.MeasureSpec.UNSPECIFIED, because this makes the most sense to me. This is a breaking change, as it makes parent size 0. Before, parent size was typically the screen size, which seems an unlikely intentional input for determining marker size and can be obtained in other ways. This change needs to be documented.

ln-12: I changed the PR accordingly. After some testing, I could not get another crash with these changes, so it seems to work.

el-qq:
@ln-12 hi. Thank you for this contribution

Please add [semantic prefixes](https://github.com/googlemaps/.github/blob/master/.github/semantic.yml) to  commits

maps-compose/src/main/java/com/google/maps/android/compose/RememberComposeBitmapDescriptor.kt
```diff
@@ -47,8 +51,8 @@ private fun renderComposableToBitmapDescriptor(
     composeView.draw(fakeCanvas)
 
     composeView.measure(
-        View.MeasureSpec.makeMeasureSpec(parent.width, View.MeasureSpec.AT_MOST),
-        View.MeasureSpec.makeMeasureSpec(parent.height, View.MeasureSpec.AT_MOST),
+        View.MeasureSpec.UNSPECIFIED,
+        View.MeasureSpec.UNSPECIFIED,
```
bubenheimer, line 55: I don't think this is correct, I believe it would need to be
```suggestion
        View.MeasureSpec.makeMeasureSpec(0, View.MeasureSpec.UNSPECIFIED),
        View.MeasureSpec.makeMeasureSpec(0, View.MeasureSpec.UNSPECIFIED),
```

Should be ok to makeMeasureSpec() just once in a top-level field.

maps-compose/src/main/java/com/google/maps/android/compose/RememberComposeBitmapDescriptor.kt
bubenheimer: > Should this be documented inline with the code or only as a release note in the changelog?

The requirement for the Composable to have a non-null size in both dimensions needs to be documented in public API KDoc documentation on `MarkerComposable()`. Also in release notes, I believe this would be a task for project maintainers, e.g. @kikoso or @dkhawk.

It could also be flagged prominently in code; personally I'd check for this case after `measure()` and throw an exception of some sort if measure returns 0 for one or both of the dimensions. This way the problem is identified in a well-defined manner if it happens with the ability to give specific instructions on what to do (set a size in the Composable).

ln-12:
I added the commit prefixes and applied all requested changes. If you are not yet happy with the wording, please let me know.

maps-compose/src/main/java/com/google/maps/android/compose/RememberComposeBitmapDescriptor.kt
```diff
@@ -39,17 +41,23 @@ private fun renderComposableToBitmapDescriptor(
...
 
-    composeView.measure(
-        View.MeasureSpec.makeMeasureSpec(parent.width, View.MeasureSpec.AT_MOST),
-        View.MeasureSpec.makeMeasureSpec(parent.height, View.MeasureSpec.AT_MOST),
-    )
+    composeView.measure(measureSpec, measureSpec)
+
+    if (composeView.measuredWidth == 0 || composeView.measuredHeight == 0) {
+        throw IllegalStateException("The ComposeView was measured to have a width or height of " +
+                "zero. Make sure the parent and content have a non-zero size.")
```
bubenheimer, line 59: According to your testing, "parent" size should no longer matter, right?

maps-compose/src/main/java/com/google/maps/android/compose/RememberComposeBitmapDescriptor.kt
bubenheimer: BTW, I meant "absence of a parent size", not "absence of a parent".

maps-compose/src/main/java/com/google/maps/android/compose/RememberComposeBitmapDescriptor.kt
ln-12: Yeah, right. I changed that.

bubenheimer (approved):

ln-12:
I was using the wrong mail for my last commit (as I was on a different machine), so the CLA check failed. Fixed that as well. Sorry for the inconvenience!

ln-12:
Hey @dkhawk, is there any open TODO on my side? We would highly appreciate getting this fix rolled out 😅

maps-compose/src/main/java/com/google/maps/android/compose/RememberComposeBitmapDescriptor.kt
ln-12: @bubenheimer Unfortunately, we now get crashes containing the exact exception from this PR. We use version `6.1.2` of this library. Do you have another idea how this could be solved? Before we forked this lib and implemented an early return which might have led to undesired results but at least did not crash.

PR #546, opened 2024-04-10
feat: Add StyleSpans support for Polylines

### LINKED ISSUE #307: Support of StyleSpan in Polylines

el-qq:
Fixes #307 🦕

There's a [similar PR](https://github.com/googlemaps/android-maps-compose/pull/308) that was proposed some time ago. Unfortunately, the author has not responded or continued working on it.
Therefore, created a similar PR that takes into account the latest changes from the comments. 

---

### DIFF
--- a/maps-compose/src/main/java/com/google/maps/android/compose/Polyline.kt
+++ b/maps-compose/src/main/java/com/google/maps/android/compose/Polyline.kt
@@ -89,7 +89,7 @@ public fun Polyline(
 }
 
 /**
- * A composable for a polyline on the map.
+ * A composable for a polyline on the map that supports a StyleSpan.
  *
  * @param points the points comprising the polyline
  * @param spans style spans for the polyline
--- a/maps-compose/src/main/java/com/google/maps/android/compose/Polyline.kt
+++ b/maps-compose/src/main/java/com/google/maps/android/compose/Polyline.kt
@@ -25,6 +25,7 @@ import com.google.android.gms.maps.model.JointType
 import com.google.android.gms.maps.model.LatLng
 import com.google.android.gms.maps.model.PatternItem
 import com.google.android.gms.maps.model.Polyline
+import com.google.android.gms.maps.model.StyleSpan
 import com.google.maps.android.ktx.addPolyline
 
 internal class PolylineNode(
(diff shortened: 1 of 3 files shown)


maps-compose/src/main/java/com/google/maps/android/compose/Polyline.kt
```diff
@@ -106,4 +213,4 @@ public fun Polyline(
             set(zIndex) { this.polyline.zIndex = it }
```
bubenheimer, line 213: I notice that all these `set()` calls should instead be `update()`, for reasons of performance & tidiness. It would be helpful to reflect this concern somewhere, but I'm not a maintainer. Could you create a separate issue for this? Or fold it into this one, but it would seem to mix different concerns.

maps-compose/src/main/java/com/google/maps/android/compose/Polyline.kt
```diff
@@ -69,12 +70,117 @@ public fun Polyline(
...
+ * @param visible the visibility of the polyline
+ * @param width the width of the polyline in screen pixels
+ * @param zIndex the z-index of the polyline
+ * @param onClick a lambda invoked when the polyline is clicked
+ */
+@Composable
+@GoogleMapComposable
+private fun PolylineImpl(
+    points: List<LatLng>,
+    spans: List<StyleSpan> = emptyList(),
```
bubenheimer, line 164: Looks like `PolylineImpl()` could be eliminated entirely in favor of having a single `Polyline()` method with a default parameter. It's a bit of a breaking change, but personally I'd favor it over having so much repetitive overload verbosity, just like with the `Marker()` overloads nightmare. I'm not a maintainer, though.

el-qq: Then there will be confusion with the color parameter - it makes no sense to specify it when we specify Spans for segments

bubenheimer (approved):

el-qq:
updated the documentation on the comments

alexandrupele:
guys thank you so much for adding this, i need it in my project. can't wait to get it released

el-qq:
@dkhawk merge? 
As I understand it, the problem with the instrumentation tests is general and outside the scope of this issue

PR #564, opened 2024-05-06
chore: switched from set() to update()

### LINKED ISSUE #547: fix:  update set() to update() for a ComposeNode (map objects)
For map objects (e.g., `Circle()`), parameters are created both in `factory` and in the `update` block

```kotlin

    ComposeNode<CircleNode, MapApplier>(
        factory = {
            val circle = mapApplier?.map?.addCircle {
                center(center)
                 //...
            } ?: error("Error adding circle")
            circle.tag = tag
            CircleNode(circle, onClick)
        },
        update = {
            update(onClick) { this.onCircleClick = it }

            set(center) { this.circle.center = it }

```

In addition, `update()` is recommended. From the description: 
```For example, use [update} when value is passed into the class constructor parameters.```

```kotlin

//....
        update = {
            update(onClick) { this.onCircleClick = it }

            update(center) { this.circle.center = it }

```

kikoso:
`update()` should be favoured over `set()`. From the documentation:

> For example, use [update} when value is passed into the class constructor parameters.

This is likely impacting performance during recomposition.

---

Fixes #547 🦕


maps-compose/src/main/java/com/google/maps/android/compose/Marker.kt
```diff
@@ -719,8 +719,8 @@ private fun AdvancedMarkerImpl(
                 }
             }
 
-            set(visible) { this.marker.isVisible = it }
-            set(zIndex) { this.marker.zIndex = it }
+            update(visible) { this.marker.isVisible = it }
+            update(zIndex) { this.marker.zIndex = it }
```
bubenheimer, line 723: I don't see these properties being set during initial AdvancedMarker construction, so this change looks invalid to me as is.

The original code here initializes the properties to default values, then "updates" them to something else via `set()`. I'd think that it would be more efficient to directly initialize them to the desired values at construction time instead. This is what is done everywhere else.

 At that point these `update()`s will do the right thing.

bubenheimer (changes requested):

maps-compose/src/main/java/com/google/maps/android/compose/Marker.kt
```diff
@@ -552,6 +556,7 @@ private fun MarkerImpl(
 @Composable
 @GoogleMapComposable
 public fun AdvancedMarker(
+    contentDescription: String? = "",
```
bubenheimer, line 559: is there a reason not to keep `state` the first parameter? Seems a better choice to me.

Also, I believe this turns the PR into a breaking change.

bubenheimer: I called it a breaking change because it will break callers that do not use named parameters.

maps-compose/src/main/java/com/google/maps/android/compose/Marker.kt
```diff
@@ -663,6 +670,18 @@ private fun AdvancedMarkerImpl(
...
+    advancedMarkerOptions.infoWindowAnchor(infoWindowAnchor.x, infoWindowAnchor.y)
+    advancedMarkerOptions.position(state.position)
+    advancedMarkerOptions.rotation(rotation)
+    advancedMarkerOptions.snippet(snippet)
+    advancedMarkerOptions.title(title)
+    advancedMarkerOptions.visible(visible)
+    advancedMarkerOptions.zIndex(zIndex)
 
     ComposeNode<MarkerNode, MapApplier>(
         factory = {
```
el-qq, line 687: maybe  should move the creation of `advancedMarkerOptions`  here?

el-qq: at recomposition will cause unnecessary initialization

maps-compose/src/main/java/com/google/maps/android/compose/Marker.kt
```diff
@@ -599,6 +745,7 @@ public fun AdvancedMarker(
 /**
  * Internal implementation for an advanced marker on a Google map.
  *
+ * @param contentDescription the content description for accessibility purposes
  * @param state the [MarkerState] to be used to control or observe the marker
  * state such as its position and info window
```
bubenheimer, line 750: Would you consider swapping the parameters here for consistency?

bubenheimer (changes requested):
Thanks for the updates. There are a few places left where state & contentDescription are still in a different order.

Also, are you still planning to deprecate the functions without contentDescription, possibly together with a tracking issue/PR? There are so many similar-looking functions in this file that it would be good not to permanently add even more.

Personally I'd just drop the old functions and make it a breaking change. It's trivial for users to change the calling params as needed.

bubenheimer:
> Regarding the breaking change: I would not like to release this as a standalone fix/feature and a breaking change. However, we can pack it together with #516 and/or #521, to minimize the amount of major releases, which are always painful.

Sounds fine. Right now the changes are inconsistent. `Marker()` has a single changed version, while `AdvancedMarker()` and `MarkerComposable()` have one changed and one unchanged version each.

dkhawk (approved):

bubenheimer (changes requested):
Please revert 450d842c38ca8ef9fb5938425408f57c3a793bb7 prior to merge, it was a bad commit.

dkhawk (approved):

bubenheimer:
> > Please revert [450d842](https://github.com/googlemaps/android-maps-compose/commit/450d842c38ca8ef9fb5938425408f57c3a793bb7) prior to merge, it was a bad commit.
> 
> This was fixed before the merge, thanks!

@kikoso @dkhawk On my end I see that https://github.com/googlemaps/android-maps-compose/commit/450d842c38ca8ef9fb5938425408f57c3a793bb7 was indeed merged to main and introduced documentation problems.

PR #591, opened 2024-07-02
feat!: added MapColorScheme

kikoso:
The current PR adds support to the [MapColorScheme](https://developers.google.com/maps/documentation/android-sdk/configure-map#map_color_scheme) released in the Maps SDK 19.0.0

This is technically a breaking change, since we are adding a new parameter to the Composable and we might have folks using named parameters. So I used the chance to move the modifier into its first position, to follow the Lint warnings and the Compose good practices guideline.

Also, adding 19.0.0 requires us to change some of our Java requirements - from 1.8 to 17


dkhawk (approved):

dkhawk (approved):

tadfisher:
This breaks lite-mode usage; see #601.

PR #595, opened 2024-07-05
chore: added kotlin-coroutines dependency

el-qq:
This PR introduces a direct dependency on the Kotlin Coroutines library, as methods from this library are used in code


dkhawk (approved):

dkhawk:
> Hi @el-qq, and thanks for the contribution. This is not required - this dependency is pulled via the `android-maps-ktx`.

I think that we generally want to avoid indirect dependencies.

PR #617, opened 2024-08-27
fix: location tracking activity to component activity

dkhawk:
Update the LocationTrackingActivity to be a ComponentActivity to make it consistent.

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,14 +22,14 @@ You no longer need to specify the Maps SDK for Android or its Utility Library as
 
 ```groovy
 dependencies {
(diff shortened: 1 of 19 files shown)


Goooler (approved):

Goooler:
And we can migrate the theme to not using AppCompat to fully get rid of AppCompat.

https://github.com/googlemaps/android-maps-compose/blob/e43b9383c8d686660a924a379bc30b180949bebe/app/src/main/AndroidManifest.xml#L28-L28

PR #618, opened 2024-08-28
chore: fully get rid of AppCompat lib

Goooler:
- It's unnecessary in a pure Compose App, we can also decrease the App size after this change.
- We don't need `com.google.android.material:material` after dropping AppCompat.

Follow up #617.


Goooler:
This has been closed accidentally, rebasing to #620.

PR #656, opened 2024-11-30
fix: update circle to be black and have a radius of 10

### LINKED ISSUE #531: Make Radius required, use Black as colour in Circle
Both Polygon and Polyline use Black as the default colour, making them visible on the Map when used without specifying a colour.
Circle by default draws a 0-radius circle. If the `radius` is specified, it still draws a `Transparent` circle on the map, making it not visible.
Update the `Circle` function to *require* the radius to be specified, and change the default colour to `Color.Black`, as used in other Composable for Maps.
As it was a simple edit, I've done it already and I would be ready to open the PR if this is agreed upon

lucanicoletti:
Fixes #531 

Update the `Circle` function to use `Color.Black` as default, and a radius of `10`

### DIFF
--- a/maps-compose/src/main/java/com/google/maps/android/compose/Circle.kt
+++ b/maps-compose/src/main/java/com/google/maps/android/compose/Circle.kt
@@ -54,8 +54,8 @@ internal class CircleNode(
 public fun Circle(
     center: LatLng,
     clickable: Boolean = false,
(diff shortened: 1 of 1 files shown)


dkhawk (approved):

lucanicoletti:
The tests are failing for `Maps API key not specified`, which shouldn't be related to this change in the specific.

PR #695, opened 2025-03-24
chore: added test for MapsInLazyColumn

kikoso:
This PR adds a text for MapsInLazyColumn, so we can track if any PR introduces a misbehavior here. As a side effect there are some visibility changes in the Activity, which is fine since it is just for demo purposes.

### DIFF
--- a/maps-app/src/androidTest/java/com/google/maps/android/compose/MapsInLazyColumnTest.kt
+++ b/maps-app/src/androidTest/java/com/google/maps/android/compose/MapsInLazyColumnTest.kt
@@ -108,6 +108,6 @@ class MapsInLazyColumnTests {
         initMaps()
         composeTestRule.onRoot().performTouchInput { swipeUp(durationMillis = 1000) }
         composeTestRule.waitForIdle()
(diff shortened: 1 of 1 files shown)


maps-app/src/main/java/com/google/maps/android/compose/MapsInLazyColumnActivity.kt
```diff
@@ -192,11 +212,13 @@ private fun MapCard(item: MapListItem, cameraPositionState: CameraPositionState)
         var map: GoogleMap? by remember { mutableStateOf(null) }
 
         fun updateIndoorLevel() {
-            activatedIndoorLevel = map!!.focusedBuilding?.run { levels.getOrNull(activeLevelIndex)?.name }
+            activatedIndoorLevel =
+                map!!.focusedBuilding?.run { levels.getOrNull(activeLevelIndex)?.name }
```
dkhawk, line 216: Nit: "!!." can just be "?." since focusedBuilding is already nullable.

dkhawk (approved):

PR #702, opened 2025-04-21
feat: added AttributionId

kikoso:
The following PR adds the AttributionId to the Maps Compose dynamically. It is being extracted from the version information in the root level Gradle file, and generated via a Gradle task.


dkhawk (approved):

wangela (approved):

matejdro:
It seems this PR has introduced a huge performance regression for apps that have `StrictMode.detectDiskReads()` enabled: https://github.com/googlemaps/android-maps-compose/issues/732

dkhawk:
> It seems this PR has introduced a huge performance regression for apps that have `StrictMode.detectDiskReads()` enabled: #732

Thank you for filing an issue.  We are looking into the issue.

PR #768, opened 2025-09-16
chore: export all demo activities and group the activities in main screen

dkhawk:
Group demo activities on main screen and export all activities in the manifest.

To make the main screen less cluttered and easier to navigate, the demo activities have been grouped into a list.
This provides a cleaner and more organized user experience.

This change also pulls in Material 3.


dkhawk:
Screenshot

image

PR #823, opened 2026-01-06
feat: Allow for info window customization on advanced markers

### LINKED ISSUE #822: Support info window customization for advanced markers
The APIs for customizing the info window are not exposed in the public API for advanced markers: 

https://github.com/googlemaps/android-maps-compose/blob/6f4b188c9a4bd834f369379e2e74e86aec8591da/maps-compose/src/main/java/com/google/maps/android/compose/Marker.kt#L832-L833

We need to be able to customize this so it would be good to expose via a public function.

pmecho:
This follows the approach used by regular markers which has separate functions for customizing just the content and customizing the whole window. Currently there is no way to customize either.

---

Fixes #822 🦕


kikoso:
Hi @pmecho . I know this is still in draft, but could you modify the PR title to follow the [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/) convention? Probably something like:

`feat: allow for info window customization on advanced markers`

pmecho:
Closing in favor of #862 which has the correct account for the CLA

PR #830, opened 2026-01-24
feat!: update Maps SDK to v20.0.0 and add internal usage attribution

dkhawk:
feat!: update Maps SDK to v20.0.0 and add internal usage attribution

Updates the Maps SDK dependencies and implements internal usage attribution using androidx.startup.

Changes:
- Replaced manual initialization with AttributionIdInitializer using androidx.startup.
- Removed GoogleMapsInitializer and MapsComposeApplication (Breaking Change).
- Updated README.md with attribution opt-out instructions.
- Added installAndLaunch Gradle task for easier demo app testing.
- Fixed ScaleBarTests density calculation.
- Updated minSdk overrides for test dependencies.

Breaking Changes:
- GoogleMapsInitializer and MapsComposeApplication have been removed. Initialization is now handled automatically by androidx.startup.
- minSdk is now 23.

### DIFF
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -37,7 +37,6 @@ junit = "4.13.2"
 junitktx = "1.3.0"
 leakcanaryAndroid = "2.14"
 mockk = "1.14.9"
-mockkAndroid = "1.14.9"
 org-jacoco-core = "0.8.14"
 jacoco-plugin = "0.2.1"
 robolectric = "4.16.1"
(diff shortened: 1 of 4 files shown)


SkylekFR:
It seems that when using the new maps compose version, the
` Fatal Exception: java.lang.NoClassDefFoundError: Failed resolution of: Lorg/apache/http/ProtocolVersion;`

has been propagated to the dependencies user.
Maybe it could have been better to fix is by declaring coreDesugaringEnabled instead of `
<uses-library
            android:name="org.apache.http.legacy"
            android:required="false" />
`
because now we need to also declares it in our own manifest

kikoso:
Hi @SkylekFR ,

This is an issue reported in maps 20.0.0, see the entry in the [Issue Tracker](https://issuetracker.google.com/issues/476788722) here.

While the issue is being fixed, would you be able to add the following to your manifest file?

```xml
<uses-library android:name="org.apache.http.legacy" android:required="false" />
```

SkylekFR:
> Hi @SkylekFR ,
> 
> This is an issue reported in maps 20.0.0, see the entry in the [Issue Tracker](https://issuetracker.google.com/issues/476788722) here.
> 
> While the issue is being fixed, would you be able to add the following to your manifest file?
> 
> ```
> <uses-library android:name="org.apache.http.legacy" android:required="false" />
> ```

I was hoping not to do it but if it's the only workaround, I will .

Thank you for the fast answer

kikoso:
> I was hoping not to do it but if it's the only workaround, I will .
> 
> Thank you for the fast answer

Not ideal, but keep an eye on the SDK issue, it might be fixed at some point in the future.

Issue #820, opened 2025-12-19
Maps: IllegalStateException when clicking "blocked Road" poi on map


toverchris:
#### Environment details

2. Android Emulator Pixel 9 Pro API 36.1
3. Library version: Release 7.0.0

#### Steps to reproduce

  1. Clone this repo and use the provided demo app.
  2. set `isTrafficEnabled = true` for google map
  3. open basic map
  4. click on blocked road poi 
#### Code example

```
    var mapProperties by remember {
        mutableStateOf(MapProperties(mapType = MapType.NORMAL, isTrafficEnabled = true))
    }
```

#### Stack trace
```
2025-12-19 11:45:40.695 20102-20102 AndroidRuntime          com.google.maps.android.compose      E  FATAL EXCEPTION: main
                                                                                                    Process: com.google.maps.android.compose, PID: 20102
                                                                                                    java.lang.IllegalStateException
                                                                                                    	at com.google.maps.api.android.lib6.common.t.d(:com.google.android.gms.policy_maps_core_dynamite@254125201@254125200025.815771884.815771884:6)
                                                                                                    	at com.google.maps.api.android.lib6.phoenix.av.a(:com.google.android.gms.policy_maps_core_dynamite@254125201@254125200025.815771884.815771884:53)
                                                                                                    	at m140.dpu.run(:com.google.android.gms.policy_maps_core_dynamite@254125201@254125200025.815771884.815771884:7)
                                                                                                    	at m140.dpw.run(:com.google.android.gms.policy_maps_core_dynamite@254125201@254125200025.815771884.815771884:88)
                                                                                                    	at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:520)
                                                                                                    	at m140.bnr$a.run(:com.google.android.gms.policy_maps_core_dynamite@254125201@254125200025.815771884.815771884:14)
                                                                                                    	at android.os.Handler.handleCallback(Handler.java:1070)
                                                                                                    	at android.os.Handler.dispatchMessage(Handler.java:125)
                                                                                                    	at android.os.Looper.dispatchMessage(Looper.java:333)
                                                                                                    	at android.os.Looper.loopOnce(Looper.java:263)
                                                                                                    	at android.os.Looper.loop(Looper.java:367)
                                                                                                    	at android.app.ActivityThread.main(ActivityThread.java:9282)
                                                                                                    	at java.lang.reflect.Method.invoke(Native Method)
[...]

PR #846, opened 2026-02-10
fix: IllegalStateException on traffic POI clicks by refactoring map listeners

kikoso:
Draft PR


dkhawk (approved):

bubenheimer:
Won't these changes reintroduce #466? They somewhat look like a revert of #478.

kikoso:
Hi @bubenheimer.  This attempts to fix #820 

With the current code structure, my reasoning is: the renderer does some internal checks when you click a traffic element; if you swap or re-register the listener at that moment (which Compose was doing during recomposition), you break the SDK's internal expectations and trigger that IllegalStateException. The MapApplier registers the listenr exactly once.

I am not yet sure of the best solution for this.

bubenheimer:
The way this code is supposed to work (as far as I remember): the only situation where Compose re-registers a listener is the special case where the listener is toggled between null and non-null. That is special-cased because the SDK treated having a "null" listener specially, with some side effect, and there was a Github issue about it for one of the listeners.

There is no re-registering of listeners otherwise; Compose swaps out implementation details (a delegate lambda) behind the scene. You think this swapping may be causing a problem here? I doubt you can get away without either swapping or re-registering the listener, as you will regress on various bugs. The code minimizes re-registering listeners because of potential performance concerns: the SDK is viewed as a black box and Compose recomposition might swap listeners constantly, which could be sluggish if the SDK made that somewhat expensive with synchronization or such.

There is a hint of concurrent execution In the stacktrace from #820; the recomposition and swapping approach was written with single-threaded main thread execution in mind, so I wouldn't rule out potential complications from that, if this is what you are thinking.

bubenheimer:
Looking at it again, the #820 stacktrace looks focused on main thread.

kikoso:
I have been thinking further about this issue. I am still trying to find the best way to get this fixed, but my understanding: traffic-managed POIs like "blocked roads" require the `OnPoiClickListener` to be stable and registered early. While the previous ComposeNode approach was designed to minimize re-registration, it still treats the listener as a dynamic entity in the composition tree. Even if a swap is just an implementation detail behind the scene, the moment the SDK sees a null or a new listener object during certain renderer states, it triggers this IllegalStateException.

bubenheimer:
The SDK would not see changes unless there is a bug or the user explicitly swaps between null and non-null listeners. A bug is certainly a possibility, the code in that area has seen many changes since the listeners redesign, and Compose has changed, too.

In any case, just trying to offer helpful pointers, I don't depend on this library anymore.

PR #881, opened 2026-03-27
perf: improve MarkerComposable rendering performance

brucemok8:
## Summary

This PR improves the rendering performance of `MarkerComposable`, particularly when displaying a larger number of markers with frequently updating content.

## Changes

- Reuse `ComposeView` instances instead of creating and disposing them for each render. The view is retained and used to generate updated bitmaps when marker keys change, and disposed only when the marker is removed.
- Update `rememberComposeBitmapDescriptor` to return a nullable type. Callers skip rendering until the bitmap is ready, preventing main thread blocking and allowing markers to appear asynchronously.

## Impact

These changes significantly reduce UI jank during marker updates. In testing, scenarios with ~25 dynamically updating markers show noticeably smoother performance.

### DIFF
--- a/maps-app/src/main/AndroidManifest.xml
+++ b/maps-app/src/main/AndroidManifest.xml
@@ -59,6 +59,9 @@
             android:name=".markerexamples.AdvancedMarkersActivity"
             android:exported="true" />
         <activity
+            android:name=".markerexamples.MarkerComposablePerformanceActivity"
+            android:exported="true" />
+        <activity
             android:name=".MapInColumnActivity"
             android:exported="true" />
         <activity
--- a/maps-app/src/main/java/com/google/maps/android/compose/Demo.kt
+++ b/maps-app/src/main/java/com/google/maps/android/compose/Demo.kt
@@ -45,6 +45,7 @@ import androidx.compose.ui.unit.dp
 import com.google.maps.android.compose.markerexamples.AdvancedMarkersActivity
 import com.google.maps.android.compose.markerexamples.MarkerClusteringActivity
 import com.google.maps.android.compose.markerexamples.MarkerClusteringRotationActivity
+import com.google.maps.android.compose.markerexamples.MarkerComposablePerformanceActivity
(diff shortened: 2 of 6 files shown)


LoyalAbbas (approved):
LGTM

kikoso:
@brucemok8 , I am not convinced about this approach. The ComposeView is now kept alive and attached to host (the window root view) for the entire lifetime of the marker, only removed when the marker leaves composition. Even with setting alpha to 0, views with alpha = 0f are still part of the root's measure/layout pass and are typically still touchable/hit-testable by default in Android. 

Also, since icon starts null and MarkerImpl is only composed once it's non-null, markers will now pop in a frame or two late instead of appearing immediately.

kikoso:
And another thought: there might be a potential race between content recomposition and bitmap capture. `setContent { currentContent() }` is only called once (inside `DisposableEffect(host))`, and relies on the nested composition's own recomposition (driven by state reads inside `currentContent()`) happening before `LaunchedEffect(*keys)` fires the actual `draw()`/bitmap capture. In the common case where the same state change drives both the content and the keys, this should resolve within the same recomposition pass since it shares the same CompositionContext/Recomposer.

brucemok8:
> @brucemok8 , I am not convinced about this approach. The ComposeView is now kept alive and attached to host (the window root view) for the entire lifetime of the marker, only removed when the marker leaves composition. Even with setting alpha to 0, views with alpha = 0f are still part of the root's measure/layout pass and are typically still touchable/hit-testable by default in Android.
> 
> Also, since icon starts null and MarkerImpl is only composed once it's non-null, markers will now pop in a frame or two late instead of appearing immediately.

The approach was a workaround we came up with at the time when adding 25 markers or updating them at the same time freezes the UI for a fair bit. The null check is to ease that and indeed it may make the marker appear a frame or two slower but we're ok showing the map first before showing the marker if it needs be, as without it it doesn't make the marker appear faster but would just freeze the UI together with the map so map and marker appear at the same time which is not ideal as well. Open to alternative approaches.

Didn't really get the potential race though.

PR #947, opened 2026-07-14
chore: update maps-ktx dependency to 6.1.0-rc03 and SDK to 37

dkhawk:
Updates dependency catalog to maps-ktx 6.1.0-rc03 and compileSdk to 37.


build.gradle.kts
```diff
@@ -49,7 +49,7 @@ val projectArtifactId by extra { project: Project ->
 allprojects {
     group = "com.google.maps.android"
     // {x-release-please-start-version}
-    version = "8.3.1"
+    version = "8.3.1-rc03"
```
kikoso: We don't need this here, right?

PR #953, opened 2026-08-04
fix: avoid re-parenting crash for MarkerInfoWindowContent/MarkerInfoWindowComposable

### LINKED ISSUE #913: MarkerInfoWindowContent not drawing content with androidx.compose:compose-bom:2026.04.01 and newer
Beginning with `androidx.compose:compose-bom:2026.04.01` setting `content` on `MarkerInfoWindowContent` has no effect (i.e. the info window stays empty).

### LINKED ISSUE #950: MarkerInfoWindowContent: app crashes when clicking on the marker on version 8.4.0
In the latest version 8.4.0,
when clicking on a `MarkerInfoWindowContent`, the app crashes.

#### Environment details

`com.google.maps.android:maps-compose version`: **8.4.0**

#### Steps to reproduce

Place a `MarkerInfoWindowContent` in the content of `GoogleMap`.
When clicking on the marker, the app crashes with this message:

> java.lang.IllegalStateException: The specified child already has a parent. You must call removeView() on the child's parent first.

#### Code example

```
    GoogleMap(
        onMapLoaded = onMapLoaded,
    ) {
        MarkerInfoWindowContent(
            state = singapore3State,
            title = "MarkerInfoWindow",
            icon = BitmapDescriptorFactory.defaultMarker(BitmapDescriptorFactory.HUE_BLUE),
        ) {
            Column(modifier = Modifier.background(Color.White).padding(5.dp)) {
                Text("Testing", color = Color.Blue, fontWeight = FontWeight.Bold)
                Text(it.title!!, color = Color.Blue)
            }
        }
    }
```

kikoso:
Fixes #950

## Summary

PR #931 fixed #913 (empty info window content on compose-ui 1.10+) by keeping the `ComposeView` permanently attached to the library's own container while the info window is shown. That works around compose-ui's attachment requirement for drawing, but it means the view handed back from `getInfoContents()`/`getInfoWindow()` already has a parent. The Maps SDK internally re-parents that view into its own default info window frame, so calling `addView()` on an already-parented view throws:

```
java.lang.IllegalStateException: The specified child already has a parent. You must call removeView() on the child's parent first.
```

This crash reproduces on device every time a `MarkerInfoWindowContent` marker is tapped, matching what is reported in #950.

## Fix

Render the Compose content into a plain `Bitmap` up front (attach, measure, layout, draw, detach), the same technique already used and proven for `MarkerComposable` icon bitmaps, and hand the Maps SDK a fresh, unattached `ImageView` wrapping that bitmap instead of the live `ComposeView`.

This keeps the compose-ui 1.10+ fix from #931 (content is fully rendered before we let go of it) while avoiding the double parent crash, since the returned view was never attached anywhere for the SDK to conflict with.

- `MapComposeViewRender.kt`: added `renderComposableToBitmap`
- `ComposeInfoWindowAdapter.kt`: rewritten to use it instead of the `RenderHandle`/`disposeForMarker` bookkeeping introduced in #931
- `MapApplier.kt`: reverted the now unneeded `disposeForMarker` wiring back to its pre-#931 shape

## Test plan

- [x] Reproduced the #950 crash on a physical device (Pixel 4) prior to this change, confirmed the exact stack trace originates from the Maps SDK's internal `addView` call
- [x] Confirmed the crash is gone after this change, including repeated open/close/reopen cycles
- [x] Confirmed `MarkerInfoWindowContent` and `MarkerInfoWindowComposable` content still render correctly under compose-bom 2026.06.01 (compose-ui 1.10+), covering the original #913 scenario
- [x] `:maps-app:connectedDebugAndroidTest` passes on device

### DIFF
--- a/maps-compose/src/main/java/com/google/maps/android/compose/ComposeInfoWindowAdapter.kt
+++ b/maps-compose/src/main/java/com/google/maps/android/compose/ComposeInfoWindowAdapter.kt
@@ -15,64 +15,55 @@
 package com.google.maps.android.compose
 
 import android.view.View
-import androidx.compose.ui.platform.ComposeView
+import android.view.ViewGroup
+import android.widget.ImageView
+import androidx.compose.runtime.Composable
 import com.google.android.gms.maps.GoogleMap
 import com.google.android.gms.maps.MapView
 import com.google.android.gms.maps.model.Marker
 
 /**
- * An InfoWindowAdapter that returns a [ComposeView] for drawing a marker's
+ * An InfoWindowAdapter that returns a [View] for drawing a marker's
  * info window.
  *
  * Note: As of version 18.0.2 of the Maps SDK, info windows are drawn by
  * creating a bitmap of the [View]s returned in the [GoogleMap.InfoWindowAdapter]
- * interface methods. The returned views are never attached to a window,
- * instead, they are drawn to a bitmap canvas. This breaks the assumption
- * [ComposeView] makes where it must eventually be attached to a window. As a
- * workaround, the contained window is temporarily attached to the MapView so
- * that the contents of the ComposeViews are rendered.
+ * interface methods. For [getInfoContents], the Maps SDK also re-parents the
+ * returned view into its own default info-window frame, so the view handed back
+ * must never already have a parent.
  *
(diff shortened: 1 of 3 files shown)


LoyalAbbas (approved):
LGTM

PR #965, opened 2026-08-14
feat: expose info window customization for advanced markers

### LINKED ISSUE #822: Support info window customization for advanced markers

kikoso:
## Summary
- `AdvancedMarkerImpl` (the private implementation backing `AdvancedMarker`) already fully supported customizing a marker's info window via `infoWindow`/`infoContent` parameters, wiring them correctly into `MarkerNode`. However, the public `AdvancedMarker` composable never exposed these parameters, so there was no way to use this from outside the library.
- Adds two new public composables, mirroring the existing pattern for regular markers (`MarkerInfoWindow` / `MarkerInfoWindowContent`):
  - `AdvancedMarkerInfoWindow` — customize the entire info window
  - `AdvancedMarkerInfoWindowContent` — customize just the info window's content, keeping the default frame
- No changes to the existing `AdvancedMarker` public API — purely additive.

Fixes #822

## Test plan
- [x] Added `testAdvancedMarkerInfoWindowContentDoesNotCrash` and `testAdvancedMarkerInfoWindowDoesNotCrash` to `GoogleMapViewTests`. Note: custom info window content is rendered to a bitmap by `ComposeInfoWindowAdapter` (not present in the Compose semantics tree), so these assert the composables wire up and `showInfoWindow()` succeeds without crashing, rather than asserting on-screen content.
- [x] Ran `GoogleMapViewTests` on a physical device (Pixel 4, API 33): all 18 tests pass.

### DIFF
--- a/maps-app/src/androidTest/java/com/google/maps/android/compose/GoogleMapViewTests.kt
+++ b/maps-app/src/androidTest/java/com/google/maps/android/compose/GoogleMapViewTests.kt
@@ -243,6 +243,49 @@ class GoogleMapViewTests {
         }
     }
 
+    @Test
+    fun testAdvancedMarkerInfoWindowContentDoesNotCrash() {
+        // Regression test for https://github.com/googlemaps/android-maps-compose/issues/822:
+        // AdvancedMarker didn't expose infoContent/infoWindow customization, even though the
+        // underlying implementation already supported it. Custom info window content is
+        // rendered to a bitmap by ComposeInfoWindowAdapter (not present in the Compose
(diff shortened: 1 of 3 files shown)


dkhawk (approved):

dkhawk:
@googlebot I fixed it.

PR #976, opened 2026-08-21
fix(maps-compose): apply mapColorScheme to GoogleMapOptions during MapView creation

dkhawk:
GoogleMap() currently applies mapColorScheme via map.setMapColorScheme() only after getMapAsync() completes. However, the Maps SDK's terrain renderer only respects FOLLOW_SYSTEM/DARK when it is present in GoogleMapOptions at MapView creation time; calling setMapColorScheme() later on an already-initialized terrain map with custom MapStyleOptions fails to trigger dark terrain shaders.

This change updates AndroidView's factory in GoogleMap.kt to automatically populate GoogleMapOptions.mapColorScheme from the mapColorScheme parameter if it has not been explicitly configured in googleMapOptionsFactory.

Also fixes GoogleMapViewTests.kt to replace vacuous tests (which asserted against a test-local variable) with real assertions against the GoogleMapOptions passed to mapViewFactory, and adds a test verifying that an explicitly configured googleMapOptionsFactory is not overwritten.

Verified via ./gradlew :maps-compose:compileDebugKotlin and test suite.


kikoso (approved):
LGTM!

PR #982, opened 2026-08-26
ci: avoid flaky wait on transient isMoving in non-animated zoom tests

kikoso:
## Summary
- `testCameraZoomIn` / `testCameraZoomOut` were intermittently failing with `ComposeTimeoutException: Condition still not satisfied after 2000 ms` while waiting for `cameraPositionState.isMoving` to become `true`.
- Root cause: for non-animated zoom (`map.moveCamera()`), the SDK fires `onCameraMoveStarted` and `onCameraIdle`/`onCameraMoveCanceled` synchronously back-to-back (see `MapUpdater.kt`), so `isMoving` can flip `true` → `false` before Compose's `waitUntil` polling ever observes the intermediate `true` state. This isn't infra flakiness, it's a race baked into the assertion order — the animated variants aren't affected because the animation genuinely holds `isMoving = true` for ~300ms.
- Removed the unobservable `waitUntil(timeout2) { isMoving }` step from the two non-animated tests; they now just wait for the move to settle (`!isMoving`) before asserting the final zoom value.

## Test plan
- [x] Ran `testCameraZoomIn`, `testCameraZoomOut`, `testCameraZoomInAnimation`, `testCameraZoomOutAnimation` on a connected emulator (Pixel 7, API 17) via `./gradlew :maps-app:connectedDebugAndroidTest`, all 4 passed.

### DIFF
--- a/maps-app/src/androidTest/java/com/google/maps/android/compose/GoogleMapViewTests.kt
+++ b/maps-app/src/androidTest/java/com/google/maps/android/compose/GoogleMapViewTests.kt
@@ -40,6 +40,7 @@ import org.junit.Rule
 import org.junit.Test
 import java.util.concurrent.CountDownLatch
 import java.util.concurrent.TimeUnit
+import kotlin.math.abs
 
 class GoogleMapViewTests {
     @get:Rule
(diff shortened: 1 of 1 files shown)


LoyalAbbas (approved):
LGTM

dkhawk (approved):

PR #1006, opened 2026-09-16
ci: move coverage history off main and wait for the emulator run

kikoso:

## 1. The history no longer lands on `main`

Branch protection on `main` uses strict required status checks, so recording coverage there charged every other contributor a branch update and a full CI re-run, emulator included.

The data now lives on the [`coverage-history`](../../tree/coverage-history) branch and `main` is never written to. Existing rows were migrated rather than restarted, so the trend stays continuous.

One clarification on the original framing, since it affects how the remaining alternatives compare. Under `strict: true` the merge commit itself already marks every open pull request out-of-date, moments before the bot commit lands. The marginal cost was therefore a second invalidation inside the roughly 3 to 10 minute window between the merge and the bot push, catching anyone who clicked "Update branch" in that gap, rather than a new class of disruption. Smaller than a full rebase bottleneck, but still a cost paid by contributors for data they did not ask for, and the fix is cheap, so the branch move is worth doing regardless.

The auxiliary branch was preferred over a daily cron because batching only reduces the invalidations rather than removing them, and it would cost per-merge granularity. Trade-off accepted: `COVERAGE.md` is no longer browsable at the repo root. `AGENTS.md` links to it, and the branch carries a `README.md` explaining what it is and why it exists.

## 2. Timing race on the instrumentation artifact

Confirmed from run history:

| Workflow | Duration | Required check |
| --- | --- | --- |
| `test.yml` | 6 to 13 min | yes |
| `instrumentation-test.yml` | 16 to 28 min | no |

On the #1000 merge itself:

```
emulator finished: 15:55:56Z
PR merged:         18:44:13Z   (2h 48m later)
bot commit:        18:47:12Z
```

That merge only recorded instrumentation because the pull request sat for nearly three hours after the emulator finished. Merged promptly, the query filtering on `conclusion == "success"` would have matched nothing and the commit would have been recorded as having no instrumentation data, permanently and silently.

The workflow now records unit coverage first, then waits for the emulator run to reach a conclusion before recording instrumentation. Unit coverage is never delayed by the emulator and is already committed if the wait times out. The wait window and poll interval are `env` values at the top of the workflow.

Polling was chosen over a `workflow_run` trigger because `workflow_run` fires on the pull request head commit, which is not the squash commit the history is keyed by, and it fires whether or not the pull request was ever merged. Idle time on an ubuntu runner is far cheaper than the macOS emulator minutes this design already avoids.

## 3. Metric clarification in COVERAGE.md

The explanation now sits near the top of the report rather than inside a section:

> **What the instrumentation number measures.** It is dominated by `maps-app`, which is the internal sample app, not a published artifact. It is *not* a direct measurement of the published `maps-compose` library. The library's code is exercised heavily by `maps-app`'s tests, but JaCoCo attributes coverage to the module that runs the tests, so it is counted under `maps-app`. [...] The unit table is the one that measures the published library modules directly.

## Concurrent push handling

Pushing twice per run made concurrent updates likely enough to matter, and a local test with two competing pushes showed the retry logic was broken. Both sides append to the end of `history.csv`, so `git pull --rebase` conflicts every time, the retry loop stalled mid-rebase, and the new rows were silently dropped.

Rows are independent records, so the correct resolution is to keep both. `push_coverage_history.sh` now redoes the update on the new tip: it takes the branch as it stands, re-applies local rows through a new `merge` subcommand keyed on `(commit, suite, module)`, regenerates the report, and pushes. The same race test then passes with both sides preserved.

## Verification

- The wait loop was extracted from the workflow and driven with a mocked `gh` across four cases: already successful, queued then in progress then successful, run failed, and no run exists. All four behave correctly.
- The concurrent-push race was reproduced against real local git repositories, the data loss confirmed, and the fix verified to preserve both sides.
- zizmor 1.25.2, the version CI pins: exit code 12 (Low), below the `> 12` gate. The only finding is `artipacked` on the data-branch checkout, which is deliberate because that checkout pushes. The `main` checkout now sets `persist-credentials: false`.
- No `${{ }}` expansion appears inside any `run:` block.

Follow-up to #1000 🦕


LoyalAbbas (approved):
LGTM

